PR #16, opened 2015-04-10
Fix docs build, was using out of date kbase-common jar.

MrCreosote:


msneddon:
builds correctly and relevant tests pass

Issue #18, opened 2015-04-13
Remove internal subsetting code (@searchable) from workspace


MrCreosote:
Currently the code is unused, increases the codebase complexity by quite a bit, and the subset data is >90% of the database size. Remove the subsetting code from the workspace - this doesn't change the API at all because it's unexposed. The plan is to use Solar/ElasticSearch for RTS, and the chance the internal subset data is actually used is basically zero.

Consider removing the type_[MD5] collections from the workspace database once this is done, as it'll drastically decrease the WS database size. Backup first, obviously.

MrCreosote:
https://github.com/kbase/workspace_deluxe/pull/19

PR #19, opened 2015-04-14
Remove internal subsetting code

MrCreosote:
See https://github.com/kbase/workspace_deluxe/issues/18

TODO: rename files where necessary


src/us/kbase/typedobj/test/files/SubsetAndMetadataExtraction/KB.spec
```diff
@@ -53,65 +48,9 @@ module KB {
...
-		list<Subdata> dl;
-		mapping<string,Subdata> dm;
-		mapping<string,list<Subdata>> dml;
-	} NestedData;
-	
-	/*
-	searchable ws_subset s i
 	@metadata ws s As String Data
 	@metadata ws i As Integer
-	@metadata ws	f	as floater
```
msneddon: Not related to subdata extraction, but I vaguely recall that extra spaces here was for testing the parser, so we should keep them.

MrCreosote: Ok, I'll add in a later PR.

src/us/kbase/typedobj/test/files/SubsetAndMetadataExtraction/KB.spec
```diff
@@ -41,8 +38,6 @@ module KB {
 	} FloatStructure;
 
 	/*
-	  @searchable ws_subset numbers
```
msneddon: Since we still have the @searchable ws_subset annotation in specs in production now, I think we should keep at least one test case with this annotation in place to ensure that having it in the parsed structure doesn't cause an error in metadata extraction.

MrCreosote: Ok, I'll add them back in in a later PR.

msneddon:
Ok, looks good.  Unit tests are passing, and the update seem to work as expected.

PR #21, opened 2015-04-14
Rename Subset classes to Metadata classes

MrCreosote:


msneddon:
just a name change refactor, still compiles fine.

PR #26, opened 2015-11-01
Initialize MongoWorkspaceDB object via dependency injection + WiredTiger compatibility

MrCreosote:
Rather than MongoWorkspaceDB initializing itself, the class now relies on the client programmer to pass in appropriately initialized dependencies, namely the BlobStore and the TypeValidator. This has several benefits:

1) MWDB now has 1 constructor with 4 parameters, as opposed to 3 with up to 8 parameters.
2) Switching the backend BlobStore type now just requires passing in the desired BlobStore, rather than changing MWDB code.
3) Since the TypeValidator is now initialized outside MWDB, and therefore so is the TypeDatabase, that means that swapping TypeVals or DBs is also as simple as changing the passed in TypeVal. For example, TypeDB could be made an interface and have 2 implementations - one that talks directly to MongoDB and one that talks to a type server via a type compiler client.
4) It is now possible to change where the MWDB internal configuration is stored. Currently it's stored in the settings collection of the mongo database. With the new changes, however, the configuration could be added to, for example, the deploy.cfg file (although I advocate leaving it alone. The fact that the configuration travels with the main workspace database is very convenient).

The drawback is that it obviously requires a little more code to initialize MWDB.

Other changes:
Split WorkspaceServer initialization into two phases, each in a different class: 
KBaseWorkspaceConfiguration, which handles the deploy.cfg configuration file contents, and InitWorkspaceServer, which builds the classes required for the WorkspaceServer class based on the configuration object. The WorkspaceServer class is now much simpler and various constants and state variables are localized to the appropriate phase of the initialization operation.
Various other minor code cleanup.


MrCreosote:
Oh yeah - this only affects internal interfaces. It's not urgent; this issue has just been bugging me for months so I decided to fix it.

MrCreosote:
Added a commit to make the codebase compatible with mongo 3+ + WiredTiger.

PR #27, opened 2015-11-09
Add bulk get permissions method

MrCreosote:
Adds a method to get permissions for multiple workspaces at once.


MrCreosote:
Mike - I think this'll be way easier to read if you merge https://github.com/kbase/workspace_deluxe/pull/26 first.

PR #29, opened 2015-12-18
Make all typespec auth declarations explicit

MrCreosote:
Authentication blocks cause bugs - too easy to move or create a function in the wrong place.

Caused get_permissions* to go from auth required -> optional.

Fixed library function to allow null users and added tests.


msneddon:
I agree that the auth blocks can be a problem.  I also find that they don't really fit in with the style of the existing syntax- if we wanted to do it right, I think, we would have something more like:

authentication required {
    funcdef function() returns ();
    ....
};

But that's obviously not a high priority.  I think in general we should encourage developers to mark every function individually for now.

PR #30, opened 2015-12-21
Workspace user manual

MrCreosote:
Far from perfect, but good enough to release and improve as needed / requested.

(temporary) built docs are here: http://dev03.berkeley.kbase.us/devWSdocs/

Might want to get @rsutormin 's eyes on this, as well as someone that uses the WS but isn't a dev. @briehl maybe?


msneddon:
This is great!  I think we can maybe even get someone more technical from the docs team to take a look as well.

Unfortunately there's a conflict after I accepted the previous PR - I'm hoping it's something trivial in the release notes or something like that.  @MrCreosote can you take a look when you get a chance?

MrCreosote:
@msneddon 

>  I think we can maybe even get someone more technical from the docs team to take a look as well.

Good idea. I leave it to you whether that should happen before the merge.

Fixed the merge conflict.

PR #31, opened 2015-12-22
Bugfix list_objects limit

MrCreosote:
https://atlassian.kbase.us/browse/WOR-180

`list_objects` is now guaranteed to return either `limit` objects or the remaining objects < `limit` that match the filters.

`skip`'s behavior is now really weird (I'll write this up when this gets merged into dev-docs or vice versa) and is deprecated. Needs to be removed anyway since it falls over when `skip` is large.

Added filtering by object ID, with which you can do pretty close to the same thing as `skip`, but much faster.


msneddon:
Even though the skip behavior is now strange, do you think existing code that iterates over objects using list and skip (and continues for several iterations even if nothing is returned) will still work?  Otherwise we'll need to deploy this fix and the data pane code in the narrative at the same time, which we should actually do soon but then will require more deployment coordination.

MrCreosote:
@msneddon 

 > Even though the skip behavior is now strange, do you think existing code that iterates over objects using list and skip (and continues for several iterations even if nothing is returned) will still work?

Well... maybe. I guess I'll just describe how things worked before and work now.

The problem is with how skip and deleted, hidden, and early versions (DHEV) of objects interact. Assume that there are 300 objects that match the list_objects filters (which exclude DHEVs), the first 200 returned by mongo are DHEV, and the limit is 100.

Previous behavior:

skip  | objects returned
---- | -----------------
0 | 0
100 | 0
101 | 1
150 | 50
200 | 100
201 | 99

Not good, obviously.

New behavior:

skip | objects returned
---- | -----------------
0 | 100
100 | 100
101 | 100
150 | 100
200 | 100
201 | 99

So a skip value <= 200 always returns the same 100 objects, which is really confusing.

So if there are no DHEVs in the objects list, there will be no difference. If there are, the behavior changes from returning < limit objects even if more are available to returning non-disjoint sets of objects for skip values that should force disjoint sets.

If the current code can handle seeing the same object more than once as it increments skip, then no problem. If not, that code will need to be fixed. This version of the WS should probably spend quite a bit of time on CI making sure things don't break.

Doing extra iterations would be unnecessary with the new code.

The only way to get skip to behave as one would expect is to implement it in the application, which wouldn't actually be too hard as a coding task but would be ungodly slow.

MrCreosote:
Also, if you're going to fix code, you might as well just fix it to use the object ID filters rather than skip.

PR #33, opened 2016-01-02
Autocomplete helper function + more

MrCreosote:
Adds a function for retrieving object names by their prefix. See timing data below.

The admin specified in the `deploy.cfg` file can no longer be removed by other admins.

Cleared the workspace database (but **not** the type DB) after every test and cleaned up some ugly code that is no longer necessary since every test starts with an empty DB.

Timing the autocomplete function
------------------------------------------

Setup - 10 workspaces each with 100K objects with names ranging from `aaaa` to `fryd`. Each timing queries all 10 workspaces with a different prefix.

I was eerily prescient when stipulating that no more than 1000 names should be returned in no more than 10ms:

    In [24]: %timeit ws.get_names_by_prefix({'workspaces': wss, 'prefix': ''});
    100 loops, best of 3: 8.25 ms per loop

    In [25]: %timeit ws.get_names_by_prefix({'workspaces': wss, 'prefix': 'a'});
    100 loops, best of 3: 9.48 ms per loop

    In [26]: %timeit ws.get_names_by_prefix({'workspaces': wss, 'prefix': 'ab'});
    100 loops, best of 3: 9.66 ms per loop

    In [27]: %timeit ws.get_names_by_prefix({'workspaces': wss, 'prefix': 'abc'});
    100 loops, best of 3: 5.29 ms per loop

    In [28]: %timeit ws.get_names_by_prefix({'workspaces': wss, 'prefix': 'abcd'});
    100 loops, best of 3: 3.96 ms per loop

It appears that return time depends on the number of names returned rather than the number of objects that match the query.


msneddon:
Looks like I waited too long :)  Accepting the docs PRs created some conflicts in this PR.  @MrCreosote can you help resolve when you get a chance?  Thanks.

MrCreosote:
@msneddon fixed

msneddon:
thanks.

PR #37, opened 2016-01-25
Workspace compatibility for Shock 0.9.6, 0.9.12-13

MrCreosote:
Workspace 0.3.5 is, to the best of my knowledge, already compatible with 0.9.6, but the tests don't pass because they use shock client features that don't work with 0.9.6. In any case, this PR makes the WS compatible with Shock 0.8.23, 0.9.6, and 0.9.12-13 (and probably the intervening versions, but they weren't tested explicitly).

Shock 0.9.8 has a fix for a bug where all of the parts files for a node are opened at the same time when closing the node. This can cause too many open files errors for large files / small chunks / low ulimit cases.

Also remove some cruft related to the old 0.0.5 workspace.


msneddon:
waiting on https://github.com/kbase/jars/pull/32

MrCreosote:
Hold off on this - 0.9.13 just came out, working on compatibility.

MrCreosote:
Ok, ready for merge as soon as https://github.com/kbase/jars/pull/32 is merged.

PR #40, opened 2016-02-02
Remove skip, deprecate list_referencing_object_counts

MrCreosote:
Skip has all sorts of bad behavior associated with it which has been
discussed ad nauseum

LROC is not really useful generally - this type of thing should be done
in the nightly metrics, if at all. Also, the current plan for GC (if we
ever get around to it) would make the current implementation break, and
alternate implementations would be much slower.

This should not be merged until the 0.4.0 release has been on next.kbase long enough that it's pretty clear it's going to production w/o changes.

### DIFF
--- a/docsource/knownuserbugs.rst
+++ b/docsource/knownuserbugs.rst
@@ -1,12 +1,5 @@
 Known user-facing bugs
 ======================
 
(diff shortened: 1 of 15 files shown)


MrCreosote:
changed target branch, see https://github.com/kbase/workspace_deluxe/pull/43

PR #43, opened 2016-02-24
Remove skip, deprecate list_referencing_object_counts

MrCreosote:
Skip has all sorts of bad behavior associated with it which has been
discussed ad nauseum

LROC is not really useful generally - this type of thing should be done
in the nightly metrics, if at all. Also, the current plan for GC (if we
ever get around to it) would make the current implementation break, and
alternate implementations would be much slower.


MrCreosote:
Hey Mike, can you merge this already?

PR #44, opened 2016-04-16
Remove external compilation code, split Workspace.java

MrCreosote:
This has been sitting in my repo for two months, time for a PR

- Removes all external compilation code. This simplifies the build process & testing quite a bit.
- Split Types.java from Workspace.java. There were 0 interconnections between the type DB code and the workspace DB code - it was basically two entirely different classes jammed into one.
- Add an interface layer between the TypeValidator and the TypeDB. This allows eventually using the TypeValidator on a system remote to the TypeDB.
- Some other minor cleanup.

@rsutormin might also want to look at this, since it affects the type DB.


msneddon:
Looks good and tests pass.

PR #47, opened 2016-05-10
Accept and emit Unix epoch times where possible

MrCreosote:
All incoming timestamps can be alternately supplied as Unix epochs.

Where possible (e.g. not tuples) epochs are provided as well as timestamps.

Also made kb-sdk compile targets.


msneddon:
Looks good, tests pass for me.

PR #50, opened 2016-05-27
Refactor get objects methods

MrCreosote:
At the API level:

* Add a new method, get_objects2, that has all the functionality of the other 4 methods that return provenance data. In particular, a user can now get only the provenance from a referenced object and can get subsets from referenced objects.
* get_object_info_new can follow references the same way get_objects2 and get_referenced_objects do.
* get_objects2 can return nulls instead of throwing an error like get_object_info_new.

At the library level:

* There's now only one getObjects type function that does everything the 4 old functions used to do. 
* The reference following logic has been pulled up from MongoWorkspaceDB to Workspace.
* Fixed an exploit where an attacker, for an arbitrary workspace, could determine the number of objects in that workspace, the number of versions of each object, and whether a particular object name exists in the workspace.


msneddon:
looks good and tests pass.

PR #54, opened 2016-06-07
Exclude objects when cloning workspaces

MrCreosote:
@briehl

https://atlassian.kbase.us/browse/WOR-233

Adds an option to exclude user specified objects when cloning workspaces. Also makes cloning workspaces slightly more efficient about excluding deleted objects.

EDIT: Added workspace html spec and some missing deprecations.


msneddon:
looks good.

MrCreosote:
Crap, this went into develop instead of dev-candidate

MrCreosote:
Mike & I agreed to leave as is. I'll merge into dev-candidate & bump the version.

PR #73, opened 2016-07-06
Fix two handle related bugs, tempdir checks

MrCreosote:
Requires https://github.com/kbase/jars/pull/43 - merged

* Saving an object with handles now requires the user to own the shock nodes (vs. read privs).
* Anonymous users can now get objects with handles without causing a null pointer exception.
* Save a temp file on startup to check that the temp dir is writeable.
* Recompiled the clients to add the autogenerated status() method.

Requires the latest versions of the handle service & manager (already in CI).


MrCreosote:
Mike, I'll ping you here when this is ready for merge.

MrCreosote:
Ok, this is ready to be merged. I'll wait until Tuesday when Keith is around before merging to develop in case something goes sideways wrt the new handle stuff.

PR #74, opened 2016-07-10
Fix bug in getObjects wrt returned data size, temp files refactor

MrCreosote:
Since 0.4.1 there are two calls to MongoWorkspaceDB.getObjects in each
Workspace.getObjects call - one for standard objects and one for object
reference paths. Each MWDB.GO call checked data limits independently,
which meant that twice the allowed data limit could be stored on disk,
in memory, and returned. 

Altered the MWDB.getObjects calls so that the set of two calls shares
the same limits.

Went down the temp files rabbit hole and reworked temp files management
in a number of places.

* Made all workspace level tests check for temp files.
* Fixed a bug where sort errors could leave temp files on disk.
* Made the server level tempfiles cleanup easier to understand and less
prone to failure.
* Fixed a bug where getObjects() errors could leave temp files on disk for similar
reasons to the data limit bug decribed above.

### DIFF
--- a/docsource/releasenotes.rst
+++ b/docsource/releasenotes.rst
@@ -38,6 +38,8 @@ UPDATED FEATURES / MAJOR BUG FIXES:
   anonymously would cause a null pointer error.
 * A temporary file is created and deleted at startup to ensure the temporary
   files directory is readable.
+* Fixed a bug where under certain circumstances more data than allowed could be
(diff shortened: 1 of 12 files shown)


MrCreosote:
@rsutormin I'm reassigning this to Mike since you're out but it wouldn't be a bad idea for you to take a look at it.

PR #88, opened 2016-10-05
Clarify message when save fails due to inaccessable handle

MrCreosote:
Also make a portion of the tests pass against the new auth service

Also also fixed the client build


msneddon:
looks good.

PR #90, opened 2016-10-25
Refactoring class names

MrCreosote:
Renamed and cleaned up a number of classes to avoid confusion between the path through the workspace reference graph and the paths into an object for the purposes of subset selection.

Went through getObjects and getObjectInformation and refactored variable names in preparation for what's to come


rsutormin:
Looks good, I'm merging it.

PR #92, opened 2016-11-02
Return resolved reference path, simplify resolving objects

MrCreosote:
* currently the OBJ_LATEST field is never set to anything but null. It was originally planned to be used for GC to show the most recent version of an object that had not been GC'd. Since the plan is now to GC all versions of an object at once, removed as many references to the field as possible without changing the db schema.
* removed the ignoreVersions argument from the base resolveObjects method. only copyOrRevert() used this argument so it's simpler to just have copyOrRevert() deal with it.
* removed a subclass of ResolvedMongoObjectID for which the features are never used.
* If the target object in a ref path is specified by a name and the name changes between resolving the ref path and getting the object, the get may fail or get the wrong object. The get now always uses the ID.
* Return the resolved reference path for an object *at the library level only*.
* Removed the mongo-specific implementation of Reference, ObjectReferenceSet, and ObjectInformation. Will continue to do the same with other classes.


MrCreosote:
@rsutormin ready for review/merge

rsutormin:
@MrCreosote: except this small comment about immutability PR looks good to me. I have only one question: can you point me to algorithm searching ref paths. I think it existed already before this PR was made, right?

MrCreosote:
@rsutormin search code is here: https://github.com/mrcreosote/workspace_deluxe/compare/5d790295cf0eb0c021b92e602fd5fd155a722a7a...dev-get_refs_BFS_expt?expand=1

But it's been almost entirely rewritten for the upcoming PR.

I'll fix the Date thing.

MrCreosote:
@rsutormin fixed the Date issue

rsutormin:
Looks good to me. Should I merge?

MrCreosote:
@rsutormin yes please

PR #93, opened 2016-11-08
Get objects without providing a reference path

MrCreosote:
Adds functionality to search up the object DAG to determine whether a user has access to an object.


src/us/kbase/workspace/database/Workspace.java
```diff
@@ -1393,38 +1427,273 @@ private ResolvedRefPaths resolveObjects(
...
+			final WorkspaceUser user,
+			final Set<ObjectIdentifier> lookup,
+			final boolean nullIfInaccessible)
+			throws WorkspaceCommunicationException, InaccessibleObjectException,
+				CorruptWorkspaceDBException {
+		if (lookup.isEmpty()) {
+			return new ResolvedRefPaths(null, null).withStandardObjects(null);
+		}
+		//could make a method to just get IDs of workspace with specific permission to save mem
+		PermissionSet pset = db.getPermissions(user, Permission.READ, false);
```
rsutormin: Does this set include globally readable (public) workspaces?

src/us/kbase/workspace/kbase/IdentifierUtils.java
```diff
@@ -267,12 +264,21 @@ private static ObjectIdentifier buildObjectIdentifier(
...
-		if (paths == null && refchain == null) {
+		final boolean searchDAG = longToBoolean(o.getFindReferencePath());
+		if (paths == null && refchain == null && !searchDAG) {
 			return oi;
 		} else if (paths == null) {
-			return new ObjectIDWithRefPath(oi, refchain);
+			if (searchDAG) {
+				return new ObjectIDWithRefPath(oi);
+			} else {
+				return new ObjectIDWithRefPath(oi, refchain);
```
rsutormin: What happens if both "find_reference_path" and one of "obj_path/obj_ref_path/to_obj_path/to_obj_ref_path" are defined. Looks like we silently disregard latter option, right?

MrCreosote: No, it's checked in earlier places in the code in the mutateObjSpecByRefString method and in the processRefChains method. At this point it's guaranteed that if searchDAG is true no paths were provided.

rsutormin:
Ok, PR looks good. I'm merging it...

PR #96, opened 2016-11-10
Referenced objects search performance timing code

MrCreosote:
merges in the timing code written for ws ver 0.5.0 - compatible with the current workspace.

Copies the old timing code & results from the experimental search ref code written for 0.3.5, which is not compatible.

Adds code for timing the 0.5.1 ws modified from the 0.3.5 code. Adds analysis based on this code


msneddon:
All of these changes are to performance testing code, which is safe to merge.

PR #97, opened 2016-11-10
Drop ref search max size to 10K based on new timings

MrCreosote:
Also fix workspace test. Not clear how this got past multiple commits.


msneddon:
relevant tests all pass

PR #98, opened 2016-11-11
 Refactor search code to simplify IO

MrCreosote:
preparation for returning paths


MrCreosote:
M & R are both out which is why I asked you, but if you don't feel comfortable merging I can wait until Monday. 

That being said, it'd be good to develop workspace expertise beyond me, M, & R, so if you want to walk through it I can do that too.

MrCreosote:
Bill authorized me to merge on his behalf, since he doesn't have write access to the ws repo

PR #99, opened 2016-11-11
Abstract reference lookups into a class

MrCreosote:


MrCreosote:
@briehl can you review this one? For some reason GH won't let me assign it to you.

PR #100, opened 2016-11-11
Move reference searching code into package & extract internal classes.

MrCreosote:
Was going to get too messy if I tried to jam all the necessary tree classes into the same class.


briehl:
Looks good.

PR #112, opened 2016-12-08
Fix bug with register_typespec_copy & updated specs

MrCreosote:
Since allTypes was a list rather than a set, it wound up with two copies
of each type when the types in the new specfile were the same as the old
specfile. In the first time through the loop, the type was removed from
typesToSave, and in the second time added to typesToUnregister. Thus any
types that are in the new spec and the old spec will be unregistered.

Changing the list to a set fixed the bug - the tests now fail on a list,
but pass with a set.

### DIFF
--- a/src/us/kbase/workspace/database/Types.java
+++ b/src/us/kbase/workspace/database/Types.java
@@ -77,9 +77,12 @@ public class Types {
 	
 	//TODO should return the spec version as well?
 	public Map<TypeDefName, TypeChange> compileNewTypeSpec(
-			final WorkspaceUser user, final String typespec,
-			final List<String> newtypes, final List<String> removeTypes,
-			final Map<String, Long> moduleVers, final boolean dryRun,
+			final WorkspaceUser user,
+			final String typespec,
+			final List<String> newtypes,
+			final List<String> removeTypes,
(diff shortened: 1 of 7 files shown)


src/us/kbase/workspace/database/Types.java
```diff
@@ -177,39 +183,51 @@ public ModuleInfo getModuleInfo(final WorkspaceUser user, final ModuleDefId modu
...
+			moduleVersionRestrictions.put(
+					localIncludedInfo.getModuleName(), localIncludedInfo.getVersionTime());
 		}
-		Set<String> prevTypes = new HashSet<String>(typedb.getModuleInfo(moduleName, lastLocalVer).getTypes().keySet());
-		Set<String> typesToSave = new HashSet<String>(extTypeSet);
-		List<String> allTypes = new ArrayList<String>(prevTypes);
+		final Set<String> prevTypes = new HashSet<>(
+				typedb.getModuleInfo(moduleName, lastLocalVer).getTypes().keySet());
+		final Set<String> typesToSave = new HashSet<>(extTypeSet);
+		final Set<String> allTypes = new HashSet<>(prevTypes);
```
MrCreosote: The bugfix is here - allTypes needs to be a Set, not a List

PR #115, opened 2016-12-13
Develop -> staging

MrCreosote:


Makefile
```diff
@@ -85,6 +85,7 @@ compile-typespec:
 		--pyclname biokbase.$(SERVICE).client \
 		--url $(URL) \
 		$(SERVICE).spec
+	rm lib/biokbase/workspace/authclient.py
```
kkellerlbl: Is something creating a bad authclient.py that this is needed?

MrCreosote: Nah, it's just not needed by anything so I delete it. Currently it's always generated by kb-sdk compile.

PR #116, opened 2016-12-14
Throw better error when user metadata contains nulls.

MrCreosote:


rsutormin:
@MrCreosote: This change looks good to me, so I'm merging it.

PR #121, opened 2017-01-28
Add un/delete workspace commands to admin interface

MrCreosote:
Also some timing code added to the performance dir.

Also also updated the initialization script to use the auth url in the deploy file it it exists when validating the backend token.


MrCreosote:
@kkellerlbl fix for init script is in this pr

PR #133, opened 2017-06-15
Remove old auth cruft

MrCreosote:
Simplify configuration and code.


MrCreosote:
@kkellerlbl note config changes. Shouldn't be anything you have to do if the current ws is running against auth2, but heads up anyway.

PR #134, opened 2017-06-15
Small updates: java 8, auth jar, deploy.cfg

MrCreosote:
Now compiles with java 8.

Update auth jar.

Move deploy.cfg -> deploy.cfg.example.


MrCreosote:
@kkellerlbl heads up

PR #136, opened 2017-06-15
Clean up test config code

MrCreosote:
Port over the auth2 system for managing test configuration rather than the insanely convoluted process that exists now.

Remove most script related documentation and test rigs (which didn't work anyway).


msneddon:
I guess I'll hold off on merging though until we figure out if anything should be updated on the deployment side of things.

kkellerlbl:
It's only going into the dev_candidate branch so it's safe to merge.  When things are merged into develop I'll watch the deployment and startup.

MrCreosote:
@kkellerlbl It's always built without the dev container - you never needed to source the env.

Note that it still deploys to /kb/deployment/services/workspace but you can override that with the TARGET variable e.g.
`make deploy TARGET=/home/kkeller/mylittlepony/explicit/workspace`

@msneddon go ahead and merge IMO. Worst case is we break CI for a few mins, and that shouldn't happen anyway for the reasons above.

kkellerlbl:
The build failed for me in the CI Dockerfile without sourcing the dev_container environment, complaining about KB_TOP not being set.

MrCreosote:
Hmm, unexpected, since KB_TOP doesn't appear in the updated makefile. I don't see where it could be complaining.

MrCreosote:
Also I've deploy many times without KB_TOP set or the env sourced, and it worked for me - what's the error?

kkellerlbl:
I got the error that KB_TOP was not set.  But this was against the develop branch, not dev_candidate, so (IIRC) KB_TOP is still referenced there.

If it's important to know the exact problem I'll need to try to build the image again; I didn't save the log from the failed build.

MrCreosote:
@kkellerlbl if you run the build in the dev container, the makefile will assume the env has been set and import the makefile common and rules files, and that error will happen. If you run the build outside the dev container, it'll work. 

Although with this PR it doesn't matter either way and the ws should be fully divorced from the dev container.

MrCreosote:
@msneddon can you merge please?

PR #158, opened 2017-08-03
TASK-895 Add prototype RESKE event handler.

MrCreosote:
Currently works for save objects.

TODO:
* [x] Add public bool to save objects event.
* [ ] Add other events (many will require changes on the RESKE side)


MrCreosote:
I'm going to go ahead and merge what exists now to keep the PR size manageable.

PR #160, opened 2017-08-16
Sort listObject output when possible, new indexes

MrCreosote:
New indexes in versions collection:
savedby
ws asc, id asc, ver dec

Listing objects will now sort by ws asc, id asc, ver dec if no filters
other than the object ID filters are specified. This makes it simpler to
list all objects in a workspace since you know if you've got all the
versions of an object without having to make another query.

Timing data shows no difference in adding the sort.

Tested with mongo explain() to be sure an index was used for sorting.

### DIFF
--- a/src/us/kbase/workspace/database/GetObjectInformationParameters.java
+++ b/src/us/kbase/workspace/database/GetObjectInformationParameters.java
@@ -6,7 +6,7 @@ import java.util.List;
 import us.kbase.typedobj.core.TypeDefId;
 
 /** Parameters for the WorkspaceDatabase getObjectInformation method.
- * Created from ListObjectsParameters.
+ * Created from {@link ListObjectsParameters}.
  * 
  * @author gaprice@lbl.gov
  *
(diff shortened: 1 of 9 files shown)


src/us/kbase/workspace/database/GetObjectInformationParameters.java
```diff
@@ -171,4 +171,19 @@ public int getLimit() {
...
 		return asAdmin;
 	}
+	
+	/** Check if there are no filters set other than the object ID filters. The object ID filters
+	 * may or may not be set.
+	 * The other filters are the two date filters, the metadata filter, the savers filter, and
+	 * the type filter.
+	 * @return true if no filters other than the object ID filters are set.
+	 */
+	public boolean isObjectIDFiltersOnly() {
```
kkellerlbl: Is the idea here just to prevent potentially bad filters?

MrCreosote: It's used elsewhere in the code to ensure that only object id filters are active before doing a sort on the ws/obj/ver numbers. Other filters being active *could* force the sort to occur in memory.

MrCreosote:
@scanon I'm going to go ahead and merge this so I can get it into the RESKE WS branch, but I still think it'd be good if you looked through it.

PR #168, opened 2017-09-12
Event handler factory for Knowledge Engine Prototype

rsutormin:


MrCreosote:
Good enough for a prototype. Needs tests before it goes to production, if it does.

PR #174, opened 2017-09-28
TASK-1095 Fix workspace clone bug

MrCreosote:


MrCreosote:
@kkellerlbl we'll need to do some index surgery before putting this up - see release notes

PR #178, opened 2017-10-17
Dockerized version of workspace

sychan:
The main codebase is unaffected, just a lot of files to support building a docker image, and bringing up a viable workspace using docker-compose.

It would be good to modify the workspace tests so that they can run against the dockerized service. I'm wondering if this would be a useful task for someone like Josh so that he gains some familiarity with the workspace?


MrCreosote:
> It would be good to modify the workspace tests so that they can run against the dockerized service. I'm wondering if this would be a useful task for someone like Josh so that he gains some familiarity with the workspace?

From our discussion, it'd make a lot more sense just to write the small integration test you wanted to ensure workspace is talking to Shock and the HS/M correctly. Running the full test suite is massive overkill.

MrCreosote:
Per our discussion @sychan is going to move as much stuff as possible into a miniKBase repo (or something) and leave the minimum amount of stuff to create the WS docker container

server_scripts/glassfish_administer_service.py
```diff
@@ -297,8 +296,9 @@ def _run_remote_command(self, *cmd):
 
 if __name__ == '__main__':
     args = _parseArgs()
+    _VERBOSE = args.verbose
```
MrCreosote: Don't make this a global variable, you philistine. Pass it into the CommandGlassfishDomain class and make it an instance variable.

sychan:
I'm going to move a bunch of stuff into a new repo and submit a new PR

PR #179, opened 2017-10-17
Dockerized

sychan:
Stripped out the mini-kbase related material and put them into kbase/mini_kb


MrCreosote:
Did a verbal review, @sychan is making some changes

sychan:
Travis stuff is working now.

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+sudo: required
+language: java
+jdk:
+  - openjdk8
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install -y ant-optional
+env:
+  global:
+    secure: "jhGBqc4vW+ALi5S2soFhBpfV11d6xXswwUA/s53WT2N4sNIA2vLOf55vI5cQcOnqOTe86PQpLXgbV5Eycp+cunaQJBTzEjAGJuHS/N1ThWyWo5MDchM78winTUYFBUrvjd1XhBx5JBJCLyQowl6yZa14LQ6uj4zbTcXK8VAin8w="
```
MrCreosote: Can you document what each of these vars are

deployment/bin/entrypoint.sh
```diff
@@ -0,0 +1,96 @@
...
+# is the dictionary context for the jinja2 template config file
+if [ "$1" ] ; then
+    if [ -r $1 ] ; then
+        DATA_SRC=$1
+    else
+        TMPDIR=/tmp/data$$
+        mkdir $TMPDIR
+        # Fetch the file, and have it error out on any redirects to avoid a 200 response
+        # that is just a redirect to a login screen
+        wget -q -nd --max-redirect=0 --no-check-certificate --header="${AUTH_DATA}" -P $TMPDIR -N $1  || \
```
MrCreosote: Why aren't we checking the certificate?

deployment/bin/entrypoint.sh
```diff
@@ -0,0 +1,96 @@
...
+if [ "$2" ] ; then
+    if [ -r $2 ]; then
+        TEMPLATE=$2
+    else
+        # Create a temp directory for downloading the file so that it is the only
+        # file there
+        TMPDIR2=/tmp/template$$
+        mkdir $TMPDIR2
+        pushd $TMPDIR2
+        wget -q -nd --max-redirect=0 --no-check-certificate --header="${AUTH_TEMPLATE}" -P $TMPDIR2 -N $2 || \
```
MrCreosote: Same re checking certificate

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM kbase/kb_jre
```
MrCreosote: Can we move this to deployment

build_docker_image.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -x
```
MrCreosote: Can we move this to deployment

push2dockerhub.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
MrCreosote: can we move this to deployment

push2dockerhub.sh
```diff
@@ -0,0 +1,32 @@
...
+# or develop branch, otherwise we don't push anything
+#
+# NOTE: IMAGE_NAME is expected to be passed in via the environment so that this
+# script can be more general
+#
+# sychan@lbl.gov
+# 8/31/2017
+
+TAG=`if [ "$TRAVIS_BRANCH" == "master" ]; then echo "latest"; else echo $TRAVIS_BRANCH ; fi`
+COMMIT=${TRAVIS_COMMIT:-`git rev-parse --short HEAD`}
```
MrCreosote: Can you explain this line with a comment for non-bashers? it's pretty opaque.

server_scripts/glassfish_administer_service.py
```diff
@@ -15,39 +15,42 @@
 
 _PARALLEL_GC = "-XX:-UseParallelGC"
 _PARALLEL_GC_ESC = "-XX\:-UseParallelGC"
+_VERBOSE = False
```
MrCreosote: Please make this an instance variable for the class, not a global variable

Dockerfile
sychan: Not really - the standard practice is to put the Dockerfile in the top level of the code repo. That is the way it is in all of the other repos.

build_docker_image.sh
sychan: The "deployment" directory is actually the /kb/deployment directory in the image, and these don't belong in the final built image. If you like I can put it into a directory where you keep administrative scripts related to the build, but I don't see one currently.

deployment/bin/entrypoint.sh
sychan: Because we may be hitting an internal server that uses self-signed certs. We don't depend on the certs for server authentication, just to encrypt traffic so that our creds don't pass in the clear.

deployment/bin/entrypoint.sh
sychan: Same as above

push2dockerhub.sh
sychan: See previous comments about what the deployment directory is for. If you like we can create a new top level directory for storing build related scripts - want to pick a name?

server_scripts/glassfish_administer_service.py
sychan: Done.

Dockerfile
MrCreosote: meh

build_docker_image.sh
MrCreosote: There isn't one. How about dockerbuild or something?

deployment/bin/entrypoint.sh
MrCreosote: Let's note that in a comment

deployment/bin/entrypoint.sh
MrCreosote: Same as above

push2dockerhub.sh
MrCreosote: See above

Dockerfile
sychan: Basically a Dockerfile is equivalent to a Makefile or a build.xml, if that helps explain things

build_docker_image.sh
sychan: Maybe just "build" and any other build related tools can go in there?

MrCreosote: Sure

sychan: Done.

push2dockerhub.sh
sychan: Done.

push2dockerhub.sh
sychan: Done

sychan:
Caught up on requested changes.

build_docker_image.sh
MrCreosote: Looks like there's 2 copies of this file now.

deployment/bin/glassfish_administer_service.py
```diff
@@ -0,0 +1,312 @@
+#!/usr/bin/env python
```
MrCreosote: Can we not duplicate this file in two places? Maybe copy from the original location to here on a build?

deployment/conf/.templates/deployment.cfg.j2
```diff
@@ -0,0 +1,37 @@
...
+auth-server-url={{ auth_server_url|default("https://ci.kbase.us/services/auth/api/legacy/KBase") }}
+# The Globus v1 authorization API url.  auth2 supports this directly
+globus-url = {{ globus_url|default("https://ci.kbase.us/services/auth/api/legacy/globus") }}
+#the credentials used for the handle manager.
+handle-manager-token={{ handle_manager_token|default("") }}
+kbase-admin-token={{ kbase_admin_token|default("") }}
+
+# configuration options for RESKE
+listeners=Search,KnowledgeEngine
+listener-Search-class=us.kbase.workspace.modules.RESKEPrototypeEventHandlerFactory
```
MrCreosote: The class was renamed, this will fail.

push2dockerhub.sh
MrCreosote: Looks like the file is still in root.

server_scripts/glassfish_administer_service.py
MrCreosote: This is still an instance variable.

deployment/conf/.templates/deployment.cfg.j2
sychan: This went in the other day as a hasty patch for testing when we pushed it into CI the other day (cut and pasted from the existing one). Tell me what the new class is called and I will make it the default in the template with an option to be overridden.

MrCreosote: https://github.com/kbase/workspace_deluxe/tree/dev-candidate/src/us/kbase/workspace/modules

sychan:
@MrCreosote I think this has all of the changes we agreed on

deployment/conf/.templates/deployment.cfg.j2
```diff
@@ -0,0 +1,38 @@
...
+temp-dir={{ temp_dir|default('ws_temp_dir') }}
+server-threads={{ server_threads|default('20') }}
+mongodb-retry={{ mongodb_retry|default('0') }}
+ignore-handle-service={{ ignore_handle_service|default("")}}
+handle-service-url={{ handle_service|default("https://ci.kbase.us/services/handleservice/") }}
+handle-manager-url={{ handle_manager|default("https://ci.kbase.us/services/handlemngr/") }}
+# The KBase authorization server url.
+auth-service-url={{ auth_service|default("https://ci.kbase.us/services/auth/api/legacy/KBase") }}
+auth-service-url-allow-insecure={{ auth_service_url_allow_insecure|default("") }}
+auth-server-url={{ auth_server_url|default("https://ci.kbase.us/services/auth/api/legacy/KBase") }}
```
MrCreosote: Delete as dicussed

PR #180, opened 2017-10-24
TASK-1161 Update search handler for new DB schema

MrCreosote:


MrCreosote:
@arfathpasha can you please review?

MrCreosote:
Merging per @sychan

arfathpasha:
Looks good to me. Only comment I have is to use the long form of the keys. The key "storageCode" requires less brain cells and time to understand than the key "strcde".

MrCreosote:
@arfathpasha true, but remember that mongo keys aren't like sql column names - those keys are stored in every one of the millions of event records that we'll be accumulating, so shortening the key names will save a lot of working set memory. In most cases the keys should be encountered in the search code itself where we have descriptive variable names for the key strings. We can do the same on the workspace side if it becomes an issue.

arfathpasha:
Ok. String constant STORAGE_CODE = "strcde" can be self documenting.

PR #185, opened 2017-11-01
Add docs to docker_image, and moved target out of build.xml into makefile

sychan:


MrCreosote:
@sychan  looks like you need to install sphinx: https://github.com/kbase/workspace_deluxe#building-documentation

PR #192, opened 2017-11-21
Make port, jvm memory size and threads configurable at container startup

sychan:
Also move default glassfish http listener port off 8080 so that workspace service can use that port if necessary (which we want in order to make container deployments consistent)

### DIFF
--- a/deployment/conf/.templates/start_workspace.sh.templ
+++ b/deployment/conf/.templates/start_workspace.sh.templ
@@ -5,7 +5,8 @@ cd /kb/deployment
 # the defaults are based on
 bin/glassfish_administer_service.py --verbose --admin $GLASSFISH/bin/asadmin --domain Workspace \
         --domain-dir /kb/deployment/services/workspace/glassfish_domain \
-        --war /kb/deployment/services/workspace/WorkspaceService.war --port {{ default .Env.service_port "7058" }} \
+        --war /kb/deployment/services/workspace/WorkspaceService.war \
+        --port {{ default .Env.service_port "7058" }} --instanceport {{ default .Env.instance_port "32768" }} \
         --threads {{ default .Env.server_threads "20" }} \
         --Xms {{ default .Env.min_memory "10000" }} --Xmx {{ default .Env.max_memory "15000" }} \
         --properties KB_DEPLOYMENT_CONFIG=$KB_DEPLOYMENT_CONFIG  && \
--- a/deployment/conf/.templates/start_workspace.sh.templ
+++ b/deployment/conf/.templates/start_workspace.sh.templ
@@ -1,0 +1,12 @@
+#!/usr/bin/env sh
+cd /kb/deployment
+
+# Values for port, threads, min/max memory are populated into source template from environment vars
+# the defaults are based on
+bin/glassfish_administer_service.py --verbose --admin $GLASSFISH/bin/asadmin --domain Workspace \
+        --domain-dir /kb/deployment/services/workspace/glassfish_domain \
+        --war /kb/deployment/services/workspace/WorkspaceService.war --port {{ default .Env.service_port "7058" }} \
+        --threads {{ default .Env.server_threads "20" }} \
+        --Xms {{ default .Env.min_memory "10000" }} --Xmx {{ default .Env.max_memory "15000" }} \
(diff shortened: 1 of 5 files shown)


deployment/conf/.templates/start_workspace.sh.templ
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
```
MrCreosote: Delete the old start file

server_scripts/glassfish_administer_service.py
```diff
@@ -76,6 +76,7 @@ def __init__(self, asadminpath, domain, domainpath, verbose):
         else:
             print('Creating domain ' + self.domain + p)
             print(self._run_local_command('create-domain', '--nopassword=true',
+                                          '--instanceport=32768',  # move instanceport off 8080
```
MrCreosote: This should be a cli parameter

deployment/conf/.templates/start_workspace.sh.templ
sychan: Done.

server_scripts/glassfish_administer_service.py
sychan: Done.

sychan:
@MrCreosote Changes made

deployment/conf/.templates/start_workspace.sh.templ
```diff
@@ -0,0 +1,12 @@
...
+cd /kb/deployment
+
+# Values for port, threads, min/max memory are populated into source template from environment vars
+# the defaults are based on
+bin/glassfish_administer_service.py --verbose --admin $GLASSFISH/bin/asadmin --domain Workspace \
```
MrCreosote: I assume you want an --instanceport argument here

server_scripts/glassfish_administer_service.py
```diff
@@ -55,11 +57,12 @@ def _parseArgs():
 
 class CommandGlassfishDomain(object):
 
-    def __init__(self, asadminpath, domain, domainpath, verbose):
+    def __init__(self, asadminpath, domain, domainpath, verbose, instanceport):
         self.asadminpath = asadminpath
         self.domain = domain
         self.path = None
         self.verbose = verbose
+        self.instanceport = instanceport
```
MrCreosote: I don't think it needs to be an instance variable, does it? It's only used in the constructor.

server_scripts/glassfish_administer_service.py
```diff
@@ -76,6 +79,7 @@ def __init__(self, asadminpath, domain, domainpath, verbose):
         else:
             print('Creating domain ' + self.domain + p)
             print(self._run_local_command('create-domain', '--nopassword=true',
+                                          '--instanceport='+str(self.instanceport),  # move instanceport off 8080
```
MrCreosote: add spaces around operators

server_scripts/glassfish_administer_service.py
```diff
@@ -35,6 +35,8 @@ def _parseArgs():
                         'glassfish/domains.')
     parser.add_argument('-p', '--port', required=True, type=int,
                         help='the port where the application runs.')
+    parser.add_argument('-i', '--instanceport', type=int, default=8080,
+                        help='port for glassfish -instanceport option in create-domain')
```
MrCreosote: missing hyphen

deployment/conf/.templates/start_workspace.sh.templ
sychan: Not really, there is a default in the glassfish admin script that just uses the normal port 8080, which is the hard coded default in glassfish.

sychan: Okay, forgot to commit that update

server_scripts/glassfish_administer_service.py
sychan: fixed

server_scripts/glassfish_administer_service.py
sychan: fixed.

server_scripts/glassfish_administer_service.py
sychan: fixed.

PR #193, opened 2017-12-01
Remove Perl deps from build

MrCreosote:
get_deploy_cfg.pm was only used to pull the service port out of
deploy.cfg to provide it to build_service_control_scripts.py in the
Makefile. However, build_service_control_scripts.py will do that itself
if no port is provided, so get_deploy_cfg.pm isn't needed.

This means no perl modules need to be installed to deploy the workspace.


MrCreosote:
@arfathpasha @tiramisu24 can you review?

MrCreosote:
@arfathpasha @tiramisu24 review request #2

MrCreosote:
This has had plenty of time for review, so I'm merging.

PR #196, opened 2017-12-06
TASK-1282: doc for local ws service

arfathpasha:


MrCreosote:
These instructions duplicate https://ci.kbase.us/services/ws/docs/buildandconfigure.html and should be merged in there

MrCreosote:
Never mind, I think this is ok as is.

docsource/developers.rst
```diff
@@ -27,6 +27,134 @@ To compile, simply run ``make compile``. The
...
+
+
+2. Start mongodb. Open a new terminal and create a directory for your mongo data (if one does not already exist), then start mongodb.
+
+.. code-block:: bash
+
+    $ mkdir ~/mongodata
+    $ mongod --dbpath ~/mongodata
+
+3. Set up the workspace for deployment in another terminal.
```
MrCreosote: Probably want to mention pulling the jars repo in a sister dir to the ws dir

docsource/developers.rst
```diff
@@ -27,6 +27,134 @@ To compile, simply run ``make compile``. The
 `kb-sdk <https://github.com/kbase/kb_sdk>`_ executable must be in the system
 path.
 
+Deploying the Worspace Service locally
```
MrCreosote: Note that this deployment uses gridFS rather than shock as a file backend and does not support handles to shock nodes in objects, and any attempt to save an object with handles will fail.

arfathpasha:
I have added the jars dependency and note on use of gridfs.

I am aware of the duplication with prod deployment. I am working to resolve this by getting one piece of documentation to reference the other (as a super set) and then just describe the delta changes needed. I think this will work if prod deployment instructions section references local deployment for the basics and then extends it with what's needed for prod. 

Also, I am looking to see how we can reduce the number of steps required, for example, by making the deploy.cfg.example file already contain the values required for a local deployment. If this sounds good to you, I can make additional PRs with these incremental improvements that remove the duplication and simplify the process for both local and prod deployment.

docsource/developers.rst
```diff
@@ -27,6 +27,140 @@ To compile, simply run ``make compile``. The
...
+--------------------------------------
+These instructions are known to work on Ubuntu 16.04 LTS.
+
+1. Install the dependencies pymongo v2.8, GlassFish v3.1.*, mongodb v3.4.* and the KBase Jars directory.
```
MrCreosote: Missed this first time - any version of mongo >= 2.6 will work

MrCreosote:
As far as further improvements to the deploy process I talked to Steve briefly and for now search is the priority. This doc is great and as long as getting the ws up for a developer is doable that's good enough.

If we did have cycles to spend I'd probably:
1) redo the make file and various scripts so you don't need to specify *RUNTIME, just use ant/asadmin/etc on the $PATH
2) clean up the javadoc warnings
3) deploy by default into a 'deploy' (or something) repo in the directory so TARGET isn't needed
4) rewrite the init script in java so pymongo (and especially a really old verison of pymongo) isn't needed
5) See if there's a way to automate the glassfish fix (maybe with a user prompt)

I don't really think that making the exmple config file specific for a dev setup is a huge help but if we did do that I'd prefer having an example config file specifically for the dev environment and then a standard file.

MrCreosote:
The other point is we'd need to coordinate with ops to make sure we don't break their deployment

arfathpasha:
Agreed. The goal ultimately should be to get any new dev to deploy locally in under 10 min and any dev to also be able to deploy in production in about the same time. 

I have changed mongo version number and added kb-sdk dependency.

docsource/developers.rst
```diff
@@ -27,6 +27,144 @@ To compile, simply run ``make compile``. The
 `kb-sdk <https://github.com/kbase/kb_sdk>`_ executable must be in the system
 path.
 
+Deploying the Workspace Service locally
+----------------------------------------
+These instructions are known to work on Ubuntu 16.04 LTS.
+
+1. Install the dependencies pymongo v2.8, GlassFish v3.1.*, mongodb >=v2.6.*, kb-sdk and the KBase Jars directory.
```
MrCreosote: ant and java8 are also dependencies, forgot again, sorry

MrCreosote: Also are you sure anything other than 3.1.2.2 will work? I seem to remember trying 3.1.1.x and failing, but I could be wrong... but to be safe I'd just spec 3.1.2.2 exactly.

arfathpasha:
No worries, added Java8, Ant and fixed Glassfish version

PR #197, opened 2017-12-18
Tomcat

sychan:


deployment/conf/.templates/setenv.sh.templ
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+#
+
+CATALINA_HOME=/usr/share/tomcat8
+
+# The first existing directory is used for JAVA_HOME (if JAVA_HOME is not
+# defined in $DEFAULT)
+JDK_DIRS="/usr/lib/jvm/java-7-openjdk-amd64 /usr/lib/jvm/jre-7-oracle-x64"
```
MrCreosote: workspace is java 8 only

Dockerfile
```diff
@@ -5,17 +5,16 @@ ARG BUILD_DATE
 ARG VCS_REF
```
MrCreosote: Needs to have glassfish vs tomcat performance comparisons across key method calls

MrCreosote: Needs documentation for how to build, deploy, and start the workspace updated. Docs need to cover starting the workspace outside of a docker container.

sychan: The plan moving forward is to test for performance issues in CI and then mitigate by adding resources, not to block the merge until the tests are run.

sychan: There used to be a docker-compose setup for how to start it using docker, however you asked that it be removed. Are you asking about how to start workspace in a docker container, or about how to start workspace under tomcat?

deployment/conf/.templates/setenv.sh.templ
sychan: I'll go strip these out - they are part of the standard Tomcat distro

Dockerfile
MrCreosote: We should just do the tests, there's no pressing need to merge

MrCreosote: > you asked that it be removed
When did I ask that?

Start with tomcat. A dev needs to have instructions for starting the same config we have in production on local metal. Currently the instructions are for glassfish.

sychan: Not merging is causing a bottleneck in the release testing. The Tomcat based container is the release candidate that we are testing in CI, which is normally built automatically by travis. Because it hasn't been merged into develop, the image with the develop tag is being built with Glassfish, and the *actual* developer container we want to use needs to be built by hand, and pushed, as well as being plowed under whenever a new develop branch is merged.

Please perform the merge - Tomcat *is* the release candidate at this point, and we are testing it before release in CI.

Also, give me a pointer to the code you used for timing tests and I will be happy to run the tests, you are the person who knows what and how to test the workspace and I know you have such code laying around.

MrCreosote: Let's just switch ci back to the develop branch then rather than rushing through an extremely significant change to the core of kbase.

The test code is in two locations - /test/performance, which is *really old* and I doubt actually works any more due to bit rot, and /performance, which is pretty old and may or may not work. I haven't used it in months / years so I'm not sure how much help I can be, but if you need help grab me tomorrow.

sychan: This is a DevOps related call as to what servlet container things get deployed in. Your concerns are noted and it looks like you don't want to participate in moving off Glassfish onto Tomcat in the manner that the rest of the DevOps teams wants to proceed.

Its all here in the this merge discussion thread: consider your hands washed of any tomcat migration related issues that may arise, I'm going to override your objections and perform the merge myself after a make a few more edits related to your request for more docs.

MrCreosote: > This is a DevOps related call as to what servlet container things get deployed in.

Why is that? It seems like a big mistake to exclude the only person who has significant expertise on the workspace and who lead the team that spent months tuning the workspace to have reasonable performance.

> Your concerns are noted and it looks like you don't want to participate in moving off Glassfish onto Tomcat in the manner that the rest of the DevOps teams wants to proceed.

I don't know where this opinion came from. I'm merely pointing out the obvious: there's no pressing need to switch, and there's there's potentially large repercussions to switching. These points, to my knowledge, have never been addressed.

sychan: I've added a bunch of documentation to docsource/buildandconfigure.rst
We can build on that moving forward.

sychan: The fact that Glassfish 3.1.2 hasn't been updated in 5 years and has about half a dozen high severity remote exploits against it is reason enough to move off ASAP. In Ops circles, running your core production service for years on an unsupported product that has had known serious exploits since 2016 would be considered culpable professional negligence. Migrating off Glassfish 3.1.2 has been a "want to have" item on a lot of people's lists for quite some time, and we should just get it done now rather than push it off.

I agree that it needs to be tested, which is why I want it into CI so we can torture it there. That is what a CI environment is for. We can rollback without much trouble, but I want to set the default to be "move forward" rather than the default being "stay put".

MrCreosote: > The fact that Glassfish 3.1.2 hasn't been updated in 5 years and has about half a dozen high severity remote exploits against it is reason enough to move off ASAP

Keith checked with security about this almost a year ago and they were ok with our setup.

> Migrating off Glassfish 3.1.2 has been a "want to have" item on a lot of people's lists for quite some time, and we should just get it done now rather than push it off.

I'm not saying don't do it. I'm saying follow good software practices and test your code before merging it into a dev-candidate branch.

PR #201, opened 2018-01-16
0.8.0-dev5 release

MrCreosote:


MrCreosote:
The main change here is that temporary narratives are ignored by the search listener.

PR #206, opened 2018-03-27
Use local auth for tests

MrCreosote:
Starts a local instance of the auth service when running tests rather than relying on test accounts in a remote auth service.

This (eventually) enables Travis testing and greatly eases testing future developments using the auth services as a roles manager for the workspace (e.g. admin / read only roles, etc).

### DIFF
--- a/src/us/kbase/workspace/test/kbase/HandleTest.java
+++ b/src/us/kbase/workspace/test/kbase/HandleTest.java
@@ -19,7 +19,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
-import org.apache.commons.lang3.StringUtils;
 import org.ini4j.Ini;
 import org.ini4j.Profile.Section;
 import org.junit.AfterClass;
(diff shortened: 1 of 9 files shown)


src/us/kbase/workspace/test/kbase/HandleTest.java
```diff
@@ -126,17 +104,35 @@ public static void setUpClass() throws Exception {
 		final String mongohost = "localhost:" + MONGO.getServerPort();
 		MongoClient mongoClient = new MongoClient(mongohost);
 
+		// set up auth
```
sychan: This seems to be setup to require local startup of the auth server, I recommend that how the auth server gets started up not be fixed into the code - so that we can startup an auth server in a separate docker container. How about we use environment variables to tell the test harness where the auth server is, and if there isn't one, *then* start one?

MrCreosote: That sounds like a good feature for another PR, but it's out of scope for this one, which is a pretty huge improvement.

sychan: How many lines of code do you think are necessary? As this is a new feature anyway, why not just put it in?

MrCreosote: Probably an hour or two to change the config, add the code, and test.

But that's not the point. The point is I did this *on the weekend* and I'm not spending any more weekend time to get it done. Be thankful you got this, which moves us towards using auth2 in a docker container by using test mode and creating tokens on the fly.

> As this is a new feature anyway, why not just put it in?
The fact it's a separate feature is a great reason to put it in a new PR.

sychan: True, you did do this on the weekend...

PR #209, opened 2018-04-02
Rename shock test to integration test

MrCreosote:
...because that's what it is


src/us/kbase/workspace/test/database/mongo/ShockBlobStoreIntegrationTest.java
```diff
@@ -44,7 +44,7 @@
 import us.kbase.workspace.database.mongo.ShockBlobStore;
 import us.kbase.workspace.database.mongo.exceptions.NoSuchBlobException;
 
-public class ShockBlobStoreTest {
+public class ShockBlobStoreIntegrationTest {
```
sychan: I think this classname is far too long, and would benefit from abbreviation to something like "ShockBlbStrIntTest"

PR #210, opened 2018-04-03
SCT-636 Fix for MIME header corruption in shock nodes

MrCreosote:
More of a workaround than a fix - just checks that the MD5 is the same
between shock and the workspace, and if not, deletes the node and tries
again.

The actual changes to the logic are pretty small; most of the PR is adding unit tests. Only integration tests existed before, which couldn't reproduce the bug.

### DIFF
--- a/docsource/releasenotes.rst
+++ b/docsource/releasenotes.rst
@@ -1,6 +1,16 @@
 Workspace service release notes
 ===============================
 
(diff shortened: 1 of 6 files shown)


scanon:
This looks okay to me.  Should we have a Java Dev review too?

PR #216, opened 2018-05-13
Replace globus url with auth2 url

MrCreosote:
First step for making ws admins based on auth2 roles

Tests pass locally


sychan:
It seems okay to me - what do you think Keith?

PR #218, opened 2018-05-15
Add AdministratorHandler interface

MrCreosote:
Allows different sources of administrators. Currently only workspace
internal admin list is supported.

Also allows for read only admin role.

Next step is implementing the interface for auth2 roles.


MrCreosote:
@sychan do you want anyone else to review this?

sychan:
My only review comment would be to add some docs about how the new admin role is supposed to work. Beyond that, it looks fine to me

MrCreosote:
The new role isn't exposed yet. I'll add docs when it can actually be used.

PR #242, opened 2019-01-24
Fix docker build

scanon:
Add missing directory


sychan:
This looks like it should work, or else be harmless. Merging it now

PR #244, opened 2019-02-02
SCT-1770 Remove Jongo overlay part 1 of many

MrCreosote:
The workspace counter stuff was especially goofy so I want to commit and
test

Also update the service checker to the python3 SDK service GET error
message


scanon:
Looks good.

PR #246, opened 2019-02-08
SCT-1770 Remove jongo overlay from save provenance

MrCreosote:


src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
```diff
@@ -1805,22 +1809,76 @@ private long incrementWorkspaceCounter(final ResolvedWorkspaceID wsidmongo,
...
 		} catch (MongoException me) {
 			throw new WorkspaceCommunicationException(
 					"There was a problem communicating with the database", me);
 		}
+		for (final ObjectSavePackage p: packages) {
+			p.provid = (ObjectId) provmap.get(p).get(Fields.MONGO_ID); // ew, side effect
+		}
+	}
+
+	private Map<String, Object> toDocument(final Provenance p) {
```
scanon: I take it that before it just used the map it had.  Now it explicitly creates the document from scratch.  Is that right?

scanon:
This looks okay as far as I can tell.  Just that one question.

src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
MrCreosote: What happened before was that Jongo had a built in Jackson object mapper that would map a POJO to BSON (I think) and hand that off to the Mongo client for processing. Since I'm ditching Jongo for various reasons previously described, I couldn't use that any more (and I actually am not a fan of it since they add a bunch of non-default settings to the mapper). I tried to more or less duplicate the mapping strategy with Jackson but couldn't get it to match after a couple hours, so I gave up and did it manually. The latter strategy is what I use in all my new services anyway since you get exactly what you want and it's explicit / no magic. I don't like magic.

scanon: Got it.  Thanks.

PR #256, opened 2019-02-20
SCT-1770 update mongo client to 3.6

MrCreosote:
* Comment out parts of a test that can't run as intended on mongo3 because the query optimizer is too smart
* Remove the KE demo listener
* Had to copy over the shock controller from java_common as it's not compatible with mongo client 3.x as is


MrCreosote:
I'll explain when we're both in the same room, too complicated otherwise.

PR #269, opened 2019-03-18
Prep for modularizing ID reference permission updates

MrCreosote:
Wrap the ID reference type in its proper class

This will be useful for a) the update to the new handle manager and b)
adding new ID types (like links directly to shock or the new data
stsore)


sychan:
@MrCreosote You lost some code coverage!

MrCreosote:
@sychan It's because it's handle related code, which is not tested in Travis because it's too hard to build the handle* services, which Tian is working on fixing right now

PR #279, opened 2019-03-26
Update documentation for @id shock annotation

MrCreosote:
Compiled with KBase SDK version 1.2.0 (commit
8def489f648a7ff5657d33ed05f41c60f4766e1b)


sychan:
I think it looks fine as the starting point. Its hard to really tell if there are any gaps in the docs until someone actually tries to use them.

PR #292, opened 2019-04-11
SCT-1855 Fix sort checking bug on embedded maps

MrCreosote:
When a map was embedded within another map (even in an array) the last
key of the embedded map would be compared with the next key of the outer
map, which could cause the data to be erroneously marked as unsorted.
When exiting a map, set the previous map key correctly based on the
known location within the JSON structure.

The IdRefTokenSequenceProviderTest tests sortedEmbeddedMap and
sortedEmbeddedMapInList failed prior to the fix. The other tests were
added after the fix was in place.


docsource/conf.py
```diff
@@ -52,7 +52,7 @@
 # The short X.Y version.
 version = '0.9'
 # The full version, including alpha/beta/rc tags.
-release = '0.9.0'
+release = '0.9.1'
```
JamesJeffryes: Should this be `-dev1`

MrCreosote: nah, not in the documentation config file.

scanon:
Looks good to me.

PR #302, opened 2019-06-14
S3 SDK V2 presigned url workaround example

MrCreosote:
Also removed BlobStore.getExternalIdentifier(). Unused and leaks implementation details.

test/s3presignedurl/S3PresignedURLExample.java isn't actually used anywhere, but it's useful code to test error handling so I'm checking it in.


MrCreosote:
Approved per @kkellerlbl

PR #304, opened 2019-06-18
Add integration tests for the S3 blob store.

MrCreosote:


MrCreosote:
By the power vested in me by @kkellerlbl, I hereby declare this PR merged.

PR #325, opened 2019-07-12
using latest hs2

Tianhao-Gu:


Tianhao-Gu:
@MrCreosote Added some config to reflect the latest HS2.

src/us/kbase/workspace/test/controllers/handle/HandleServiceController.java
```diff
@@ -140,6 +140,11 @@ private File createHandleServiceDeployCfg(
 		hs.add("mongo-host", "127.0.0.1");
 		hs.add("mongo-port", "" + mongo.getServerPort());
 		hs.add("mongo-database", mongoDB);
+		hs.add("mongo-user", "");
+		hs.add("mongo-password", "");
+
+		hs.add("start-local-mongo", 1);
```
MrCreosote: What does this mean?

Tianhao-Gu: old HS has SQL started locally. I had an impression that mongo server will start locally as well. But Keith said mongo will be remotely hosted. So I added this config if set, mongo will start locally. Otherwise, it won't start local mongo.

MrCreosote: But the tests already have a mongod running at this point - the passed in MongoController manages that, and that's the server that mongo.getServerPort() returns. Another mongo server shouldn't be necessary.

MrCreosote:
I'm seeing this in the tests: `Could not establish a connection to the Handle Service at http://localhost:43063: "invalid literal for int() with base 10: 'FAKEHANDLE_-100'"`

src/us/kbase/workspace/test/controllers/handle/HandleServiceController.java
Tianhao-Gu: 👍

src/us/kbase/workspace/kbase/InitWorkspaceServer.java
```diff
@@ -516,7 +516,7 @@ private static AbstractHandleClient getHandleServiceClient(
 						"https is recommended.");
 				cli.setIsInsecureHttpConnectionAllowed(true);
 			}
-			cli.isOwner(Arrays.asList("FAKEHANDLE_-100"));
```
Tianhao-Gu: @MrCreosote HS2 is attempting to search for number after 'KBH'. In this case, it does a `int('FAKEHANDLE_-100')` and errors. Should I make it ok to pass any arbitrary handle_id?

MrCreosote: How does the old service handle it?

Tianhao-Gu: seems like it's just return false

Tianhao-Gu:
@MrCreosote could you take another look?

src/us/kbase/workspace/kbase/InitWorkspaceServer.java
MrCreosote: So if it sees a namespace it doesn't recognize it returns false?

Actually, here's the code: https://github.com/kbase/handle_service/blob/master/lib/Bio/KBase/AbstractHandle/AbstractHandleImpl.pm#L1311-L1320

So it'll try and strip off the namespace, fail to do so, and then not find anything.

That's not a great API in my opinion, but now's not the time to try and fix it. What does your code do?

MrCreosote: Or maybe now is the time to try and fix it. HS2 could throw an error and we could either check for that error and consider it a successful start or we could pass another handle ID where it's reasonable to return false (like KBH_-1 or something)

Tianhao-Gu: @MrCreosote I updated HS2 to return false if handle_id is not found in the DB.

MrCreosote: But the DB only has the number part of the handle ID, right? so if KBH_42 exists and I send FAKE_42 will that return true?

Tianhao-Gu: @MrCreosote no. it is doing a split('KBH_'). 'FAKE_42' will return false.

MrCreosote: Ok, so it does the same thing as the old service then? It's searching a set of integers for 'FAKE_42'?

Tianhao-Gu: yup. yup.

MrCreosote: OK. I can live with that but we have a chance to make a better interface, I think. Up to you.

Tianhao-Gu: cool. I think I am pretty happy with what it is now. I think it's reasonable to return false for `isOwner` if it's a fake handle id.

PR #326, opened 2019-07-23
Proposed entrypoint change

scanon:
Switch entrypoint to use a script and use environment variables to set the dockerize parameters.


scanon:
So the compose would look something like...

  ws:
    image: kbase/workspace_deluxe
    environment:
      CONF_URL: https://gitlab.kbase.lbl.gov:1443/devops/serviceconf/raw/master/workspace/workspace-prod.yml
   ....

PR #332, opened 2019-08-17
Fix NPE on old objects without external IDs field

MrCreosote:
New test fails with an NPE without the new NPE checking code


MrCreosote:
Merging once tests pass since this is basically identical to #331

PR #339, opened 2019-10-01
Make tests pass locally & with java 11

MrCreosote:
Added a Pipfile for easily creating a HS2 test environment.


MrCreosote:
Merging per @scanon

PR #341, opened 2019-10-03
Add workspace events to Kafka notifier

MrCreosote:
also some minor documentation fixes

Being a little bit lazy here and relying on unit tests. Once it's up in CI we can check if Kafka is getting the events.

### DIFF
--- a/src/us/kbase/workspace/test/modules/KafkaNotifierFactoryTest.java
+++ b/src/us/kbase/workspace/test/modules/KafkaNotifierFactoryTest.java
@@ -3,9 +3,7 @@ package us.kbase.workspace.test.modules;
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
-import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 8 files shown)


src/us/kbase/workspace/test/modules/KafkaNotifierFactoryTest.java
```diff
@@ -82,64 +80,194 @@ private TestMocks(
...
+						.build())))
+				.thenReturn(fut);
+
 		mocks.listener.setWorkspaceMetadata(
-				new WorkspaceUser("u"), 1L, Instant.ofEpochMilli(10000));
+				new WorkspaceUser("user7"), 23L, Instant.ofEpochMilli(40000));
 		
-		verify(mocks.client, never()).send(any());
+		verify(mocks.client).partitionsFor("mytopic3");
+		verify(fut).get(35000, TimeUnit.MILLISECONDS);
```
jayrbolton: Wondering how this works as this syntax is pretty confusing to me. I'm assuming you're checking that you get a message that matches the Map you create above. But it also seems like you are publishing that map to "mytopic32"? That doesn't make sense to me. What is the meaning of `verify(fut).get(35000, TimeUnit.MILLISECONDS)`?

MrCreosote: yeah, full mocks often throw people for a loop the first time they see them. What's happening is that in `initTestMocks` a mock object for the Kafka client is created and the Kafka notifier is initialized with that mock client. The `when().thenReturn()` stanza is similar to your mock server - when the mock client sees the `send()` call in the `when()` call, it returns the object in the `thenReturn()` call. In this case it returns a mock `Future` object, and the expected call is verified on that object.

so the `when()...` verifies the call I expect to the client happened.

MrCreosote: I forgot to mention, of course, that the `setWorkspaceMetadata` call on the Kafka notifier sets the whole process in motion. The `when` needs to be set up before that happens, and the `verifies` occur after.

MrCreosote: Reading this stuff becomes second nature pretty fast

jayrbolton: I guess what makes it confusing is trying to think about what the "fut" object actually is or does. I'm guessing it's kind of like a push-pull socket.

MrCreosote: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html

MrCreosote: In this case, it's just a mock returned by the mock kafka client, and the test just verifies that the expected method was called on it with the expected args.

PR #342, opened 2019-10-04
Stop spamming handle service logs

MrCreosote:
Now there's a status method, can just use that

Also move the check method into the server class. No point in putting it
in the methods class.

### DIFF
--- a/src/us/kbase/workspace/kbase/WorkspaceServerMethods.java
+++ b/src/us/kbase/workspace/kbase/WorkspaceServerMethods.java
@@ -16,7 +16,6 @@ import static us.kbase.workspace.kbase.IdentifierUtils.processWorkspaceIdentifie
 import static us.kbase.workspace.kbase.KBasePermissions.translatePermission;
 
 import java.io.IOException;
-import java.net.URL;
 import java.net.UnknownHostException;
 import java.text.ParseException;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


src/us/kbase/workspace/kbase/WorkspaceServerMethods.java
```diff
@@ -96,45 +92,24 @@
...
 		this.idFacBuilder = idFacBuilder;
-		this.handleServiceUrl = handleServiceUrl;
 		this.auth = auth;
 	}
 	
 	public ConfigurableAuthService getAuth() {
 		return auth;
 	}
 	
-	public URL getHandleServiceURL() {
```
scanon: Why was this removed?  Where is that call now?

scanon: Maybe I get it... There is still a method in the config part.

MrCreosote: the handle service stuff living in that class was cruft from older implementations - there was no real reason for it to be there, and it made more sense just to move all that code to the main server class.

MrCreosote: And yeah, the main server class is now getting a client from the object returned by `initWorkspaceService` or whatever it's called, rather than pulling the url from this class.

PR #348, opened 2020-03-23
Sync develop -> dev-candiate

MrCreosote:


MrCreosote:
@sychan may want to take a look, but merging since this is mostly trivial

PR #351, opened 2020-06-19
Add ErrorOr class

MrCreosote:
This class will be used to return either an object, such as a Workspace
object, or an error code that explains why the object could not be
retrieved. The error code will be exposed in API methods that use the
ignoreError parameter rather than just a null value.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -62,7 +62,6 @@
     <include name="apache_commons/commons-io-2.4.jar"/>
     <include name="apache_commons/commons-lang3-3.1.jar"/>
     <include name="mongo/mongo-java-driver-3.8.2.jar"/>
-    <include name="bson4jackson/bson4jackson-2.2.0-2.2.0.jar"/>
     <include name="slf4j/slf4j-api-1.7.7.jar"/>
     <include name="logback/logback-core-1.1.2.jar"/>
     <include name="logback/logback-classic-1.1.2.jar"/>
--- a/src/us/kbase/workspace/database/exceptions/ErrorOr.java
+++ b/src/us/kbase/workspace/database/exceptions/ErrorOr.java
@@ -1,0 +1,96 @@
+package us.kbase.workspace.database.exceptions;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.NoSuchElementException;
+import java.util.Optional;
+
+/** 
+ * A class that represents an object or an error that occurred attempting to create or retrieve
+ * the object.
+ */
+public class ErrorOr<T> {
+
+	// TODO ERROR may want to support including custom error fields? Different errors = different fields
+	// TODO LOMBOK
+	// TODO CODE return this from bulk methods rather than object / null 
+	
+	private final ErrorType error;
+	private final T object;
+	
+	
+	/** Create an object containing an error.
+	 * @param error the error.
+	 */
+	public ErrorOr(final ErrorType error) {
+		this.object = null;
+		this.error = requireNonNull(error, "error");
+	}
+	
+	/** Create an object containing a result object. Null is acceptable, but consider using
+	 * {@link Optional} instead.
+	 * @param object the object.
+	 */
+	public ErrorOr(final T object) {
+		this.object = object;
+		this.error = null;
+	}
+	
+	/** Returns true if this instance contains an error, false if an object.
+	 * @return true if this instance contains an error.
+	 */
+	public boolean isError() {
+		return error != null;
+	}
+	
+	/** Get the error, if any.
+	 * @return the error.
+	 * @throws NoSuchElementException if this instance does not contain an error.
+	 */
+	public ErrorType getError() {
+		if (error == null) {
+			throw new NoSuchElementException("error");
+		}
+		return error;
+	}
+	
+	/** Get the object, if any. The object may be null.
+	 * @return the object.
+	 * @throws NoSuchElementException if this instance does not contain an error.
+	 */
+	public T getObject() {
+		if (error != null) {
+			throw new NoSuchElementException("object");
+		}
+		return object;
+	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ((error == null) ? 0 : error.hashCode());
+		result = prime * result + ((object == null) ? 0 : object.hashCode());
+		return result;
+	}
+
+	@Override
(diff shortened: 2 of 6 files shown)


MrCreosote:
ugh, apt update errors

MrCreosote:
@jsfillman Looks like LGTM isn't working correclty here, it's not detecting the java code

build.xml
```diff
@@ -62,7 +62,6 @@
     <include name="apache_commons/commons-io-2.4.jar"/>
     <include name="apache_commons/commons-lang3-3.1.jar"/>
     <include name="mongo/mongo-java-driver-3.8.2.jar"/>
-    <include name="bson4jackson/bson4jackson-2.2.0-2.2.0.jar"/>
```
xpe, line 65: Did this sneak in from another cause/change? (That's my guess)

src/us/kbase/workspace/database/exceptions/ErrorOr.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Optional;
+
+/** 
+ * A class that represents an object or an error that occurred attempting to create or retrieve
+ * the object.
+ */
+public class ErrorOr<T> {
+
+	// TODO ERROR may want to support including custom error fields? Different errors = different fields
+	// TODO LOMBOK
```
xpe, line 15: Lombok = https://projectlombok.org/ ?

xpe:
From Gavin:

> There are workspace methods that return multiple things. By default if getting one thing fails (authorization, deleted, no such id etc) the whole call fails. You can ask for partial results but the things are just missing from the results. This is step 1 of many of also returning why you can't get the thing, which is basically required when doing integrations with other systems, which currently are a lot harder than they need to be

src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
```diff
@@ -1014,6 +1014,7 @@ public ResolvedWorkspaceID resolveWorkspace(
 			final Set<WorkspaceIdentifier> wsis,
 			final boolean allowDeleted,
 			final boolean allowMissing)
+			// TODO CODE don't think comm exception needs to be checked. Usually will just fail.
```
xpe, line 1017: Can you add a test for this?

src/us/kbase/workspace/database/exceptions/ErrorOr.java
```diff
@@ -0,0 +1,96 @@
...
+	// TODO CODE return this from bulk methods rather than object / null 
+	
+	private final ErrorType error;
```
xpe, line 18: I can see that coupling `ErrorType` with `ErrorTo` here might be pragmatic... however, I could see `ErrorOr` becoming something we use more broadly, in which case we'd want the error type to be a second generic.

build.xml
MrCreosote: dependency tree cruft that was missed when removing the root jar in the tree

src/us/kbase/workspace/database/exceptions/ErrorOr.java
MrCreosote: correct

src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
MrCreosote: Not sure what you mean, this is just a comment. There's nothing to test.

Maybe you mean changing the communication exception to an unchecked exception when and if that happens? There's no way to test for that either, though...

src/us/kbase/workspace/database/exceptions/ErrorOr.java
MrCreosote: That could make sense, but in the case of the workspace not necessary, so I'm pulling a YAGNI here.

build.xml
MrCreosote: Really need to move to gradle at some point

src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
xpe: > TODO CODE don't think comm exception needs to be checked. Usually will just fail.

When I see a TODO comment, I try to ask "What is actionable here?" with the hopes of moving it to a ticketing system. For this one, I'd also ask how we can more certainty around the "I don't think" part.

src/us/kbase/workspace/database/exceptions/ErrorOr.java
xpe: Yes, we are saying the same thing in this case.

src/us/kbase/workspace/database/exceptions/ErrorOr.java
xpe: I'd like to move TODOs out of the code into tickets.

build.xml
xpe: 1. If you get a chance, splitting this change out into its own commit helps.

2. What's the next step re: Gradle: create a ticket?

MrCreosote: For 1): yeaaaah... I'm not going to spend my time on that. Sorry.
For 2): Nothing, I'll get to it when I get to it. It's not something that needs to be tracked.

src/us/kbase/workspace/database/exceptions/ErrorOr.java
MrCreosote: It seems you and I are pretty far apart on how we treat TODOs. I'm happy to have that discussion, but not in a PR review - we should probably move it to slack or something.

src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
MrCreosote: See above re TODOs. Re certainty, I need to look at how the exception is used in the code base, which is not something I'm going to get to immediately and is not important enough for a ticket.

src/us/kbase/workspace/database/exceptions/ErrorOr.java
xpe: We can discuss over a video chat.

MrCreosote:
There haven't been any new comments in several days, so I'm going to go ahead and merge.

xpe:
I didn't approve the PR. You can find someone else if you want an approval.

PR #352, opened 2020-07-17
Clean up Handle ID handler factory initialization

MrCreosote:
Looks like cruft leftover from prior to the handle service / handle
manager merge

This class will be used as the starting point for the Sample service
integration, so cleanliness is important


MrCreosote:
Loss of coverage is due to fewer lines of code. The diff coverage is no worse than it was before.

sychan:
Looks pretty benign, ok to merge.

PR #361, opened 2020-08-20
Add sample service controller

MrCreosote:
Writing a main() test method would require bringing up arango,
auth, and the workspace. Too much of a pain, tested in the integration
test rig that is currently local WIP.

### DIFF
--- a/src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
+++ b/src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
@@ -1,0 +1,229 @@
+package us.kbase.workspace.test.controllers.sample;
+
+import static us.kbase.common.test.controllers.ControllerCommon.findFreePort;
+import static us.kbase.common.test.controllers.ControllerCommon.makeTempDirs;
+import static us.kbase.workspace.database.Util.checkString;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.LinkedList;
+import java.util.Map;
+
+import org.apache.commons.io.FileUtils;
+import org.ini4j.Ini;
+import org.ini4j.Profile.Section;
+
+import us.kbase.workspace.test.controllers.arango.ArangoController;
+
+/** Q&D Utility to run the Samnple Service for the purposes of testing from Java.
+ * @author gaprice@lbl.gov
+ *
+ */
+public class SampleServiceController {
+	
+	private static final String SAMPLE_SERVICE_NAME = "SampleService";
+	
+	private final Process sampleService;
+	private final int port;
+	private final Path tempDir;
+	private final Path logfile;
+	
+	/** A container for a multitude of arangoDB parameters required by the Sample Service
+	 * controller.
+	 *
+	 */
+	public static class SampleServiceArangoParameters {
+		// ugh.
+		public final String arangoUser;
+		public final String arangoPwd;
+		public final String arangoDB;
+		public final String sampleCollection;
+		public final String versionCollection;
+		public final String versionEdgeCollection;
+		public final String nodeCollection;
+		public final String nodeEdgeCollection;
+		public final String dataLinkCollection;
+		public final String workspaceObjectVersionShadowCollection;
+		public final String schemaCollection;
+		
+		/** Create the parameter container.
+		 * @param arangoUser the ArangoDB user account to use to connect to ArangoDB.
+		 * @param arangoPwd the user's password.
+		 * @param arangoDB the database to connect to.
+		 * @param sampleCollection the collection in which to store sample information.
+		 * @param versionCollection the collection in which to store version information.
+		 * @param versionEdgeCollection the collection in which to store version edge information.
+		 * @param nodeCollection the collection in which to store node information.
+		 * @param nodeEdgeCollection the collection in which to store node edge information.
+		 * @param dataLinkCollection the collection in which to store data link information.
+		 * @param workspaceObjectVersionShadowCollection the collection in which to store
+		 * workspace shadow object information.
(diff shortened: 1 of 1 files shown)


src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.LinkedList;
+import java.util.Map;
+
+import org.apache.commons.io.FileUtils;
+import org.ini4j.Ini;
+import org.ini4j.Profile.Section;
+
+import us.kbase.workspace.test.controllers.arango.ArangoController;
+
+/** Q&D Utility to run the Samnple Service for the purposes of testing from Java.
```
MrCreosote, line 22: Fixed typo locally, will be in a future PR

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
```diff
@@ -0,0 +1,229 @@
...
+				final String versionEdgeCollection,
+				final String nodeCollection,
+				final String nodeEdgeCollection,
+				final String dataLinkCollection,
+				final String workspaceObjectVersionShadowCollection,
+				final String schemaCollection) {
+			this.arangoUser = checkString(arangoUser, "arangoUser");
+			this.arangoPwd = checkString(arangoPwd, "arangoPwd");
+			this.arangoDB = checkString(arangoDB, "arangoDB");
+			this.sampleCollection = checkString(sampleCollection, "ampleCollection");
```
briehl, line 82: string should be "sampleCollection" ?

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
```diff
@@ -0,0 +1,229 @@
...
+				.redirectErrorStream(true)
+				.redirectOutput(logfile.toFile());
+		Map<String, String> env = samplepb.environment();
+		env.put("KB_DEPLOYMENT_CONFIG", ssIniFile.getAbsolutePath().toString());
+		env.put("KB_SERVICE_NAME", SAMPLE_SERVICE_NAME);
+		env.put("PYTHONPATH", libDirPath);
+		samplepb.directory(new File(libDirPath));
+		sampleService = samplepb.start();
+
+		Thread.sleep(1000); //let the service start up
```
briehl, line 158: Will there be a problem if it doesn't start up in this time? Anything that should be trapped?

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
```diff
@@ -0,0 +1,229 @@
...
+	
+	public void destroy(final boolean deleteTempFiles, final boolean dumpLogToStdOut)
+			throws IOException {
+		if (sampleService != null) {
+			sampleService.destroy();
+		}
+		if (dumpLogToStdOut) {
+			try (final BufferedReader is = Files.newBufferedReader(logfile)) {
+				is.lines().forEach(l -> System.out.println(l));
+			}
```
briehl, line 223: I'm a little out of the loop with modern Java-isms. Is a `try` without a `catch` allowed these days, and just propagates the uncaught error out to the next layer in the call stack? Is the try really needed here since you're just throwing `IOException`s anyway?

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
MrCreosote: Fixed this locally, will be in a future commit.

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
MrCreosote: The tests will fail with the workspace complaining it can't contact the sample service. I've never seen this happen, 1s appears to be more than enough time.

> Anything that should be trapped?

I don't think so. If this wasn't a test rig my opinion would be different, of course.

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
MrCreosote: `try` without a `catch` is the equivalent of python `with`. In this case it ensures the `BufferedReader is` is closed when the `try` block exits.

MrCreosote: Or more accurately, `try` with parenthesis around a closeable resource is the equivalent of python `with`. Whether there's a `catch` or not is immaterial.

MrCreosote: https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
briehl: Why not this PR?

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
briehl: 👍

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
briehl: Gotcha. LGTM.

src/us/kbase/workspace/test/controllers/sample/SampleServiceController.java
MrCreosote: Because locally there's 21 commits sitting on top of this PR and I don't want to go through the annoyance of making a new branch, fixing the message, and then cherry picking all the other commits on top of that branch for such a trivial change

MrCreosote: If it actually affected functionality then sure

PR #364, opened 2020-08-25
Add arango and sample service to test configuration

MrCreosote:
Updates the test config example file for sample service configuration items and the test configuration helper code to pull said items.


src/us/kbase/common/test/TestCommon.java
```diff
@@ -157,7 +168,11 @@ public static String getShockVersion() {
 	}
 	
 	public static String getHandleServiceDir() {
-		return getTestProperty(HANDLE_SERVICE_DIR, true);
+		return getTestProperty(HANDLE_SERVICE_DIR);
```
MrCreosote, line 171: This is a requirement now, don't allow nulls

PR #367, opened 2020-09-03
new S3-specific options

kkellerlbl:
Add new S3-specific config options to deployment template.


MrCreosote (approved):
LGTM. Sloppy on my part though.

PR #374, opened 2020-09-22
Add dependency checking code to the ID handlers

MrCreosote:
Makes much more sense and will allow simplifying the WorkspaceServer and
the InitWorkspace classes. Will also make it easier to add more ID types
in future as changes to those classes will no longer be necessary.

Next: integrate into the rest of the server and remove dependency checking code from the WorkspaceServer class.


briehl (approved):
Seems like a better pattern that'll be easier to implement going forward, if there's new handle types. Tests seem to cover it as well.

If there's no other cases, I'd say this is good.

PR #375, opened 2020-09-26
Update the S3 jars.

MrCreosote:
Notes:
The behavior of the expiration time changed with the new clients and
caused all S3 saves to fail, so for now I removed the expiration time.
This means the expiration time will default to a week:
https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/auth/signer/params/Aws4PresignerParams.Builder.html#expirationTime-java.time.Instant-
Since the URL with the creds only lives in the workspace memory and the
reference is released as soon as the object is saved, this doesn't seem
like a huge problem, as the intent is to replace the presign code next.

This PR is branching off of dev-candidate slightly after the last merge
to develop to pick up the Jackson jars updates. However, it's branched
*prior* to any sample service changes so this can be merged without
bringing in all that code if necessary.


bio-boris (approved):
Seems reasonable to me

PR #376, opened 2020-09-28
Replace S3 presign implementation

MrCreosote:
The updated S3 client now has built in presign functionality, use that
instead of the prior workaround
Also restore the 15m expiration time for signed URLs

### DIFF
--- a/src/us/kbase/workspace/database/mongo/S3ClientWithPresign.java
+++ b/src/us/kbase/workspace/database/mongo/S3ClientWithPresign.java
@@ -8,10 +8,10 @@ import java.io.DataInputStream;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPut;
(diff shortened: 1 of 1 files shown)


MrCreosote:
Note: the original glassfish server timeouts were set to 15m by the glassfish startup script in the workspace repo. Apparently tomcat, which is the server we're using now, has 30m timeouts by default: https://www.codejava.net/java-ee/servlet/how-to-set-session-timeout-for-java-web-application

Maybe just set the url expiry to a couple hours to be safe.

jsfillman (approved):
I don't have 100% context, but the changes make sense as far as what code was removed and what changes replaced it.

src/us/kbase/workspace/database/mongo/S3ClientWithPresign.java
```diff
@@ -131,17 +147,21 @@ public void presignAndPutObject(
...
-				final byte[] buffer = new byte[1000];
-				try (final InputStream in = res.getEntity().getContent()) {
-					new DataInputStream(in).readFully(buffer);
-				} catch (EOFException e) {
-					// do nothing
+			try {
+				if (res.getStatusLine().getStatusCode() > 399) {
+					final byte[] buffer = new byte[1000];
+					try (final InputStream in = res.getEntity().getContent()) {
+						new DataInputStream(in).readFully(buffer);
```
MrCreosote, line 154: Noat that line 158 is covered by tests, which means that 154 must be covered by tests, since the buffer is used there

PR #382, opened 2020-09-30
Make IDReferenceType comparable, add tests

MrCreosote:
Also minor clean up, documentation


briehl (approved):
Simple enough, expands the IdReferenceType to have a `compareTo` function 👍

PR #383, opened 2020-09-30
Get dependency status from the ID handlers vs. custom methods

MrCreosote:
Builds on the work from the past few commits.

Tests for `WorkspaceServerMethods` at this point are all integration tests that start at the JSON RPC API. Fixing that is waaaay out of scope for this PR. Nonetheless, the changes are tested.

Next step: Delete all the old dependency status code which is now cruft.


sychan (approved):
No funny smells...

PR #386, opened 2020-10-12
Set release date for 0.11.3 in notes

MrCreosote:


src/us/kbase/common/test/service/ServiceCheckerTest.java
```diff
@@ -54,7 +54,7 @@ public static void beforeClass() throws Exception {
 		TESTS.put("short200", new TestSpec(
 				"URL " + KBASE_ENV + "/services/shock-api does not point " +
 				"to a KBase SDK generated service. Code: 200, message: OK, " +
-				"content: {\"attribute_indexes\"",
+				"content: {",
```
kkellerlbl, line 57: Is this diff important?

kkellerlbl (approved):
Approved modulo one comment (which looks harmless).

src/us/kbase/common/test/service/ServiceCheckerTest.java
MrCreosote: It allows the tests to pass. `attribute_indexes` is shock only, the blobstore doesn't have that field.

kkellerlbl: :+1:

PR #388, opened 2020-10-15
Clear DBs between tests

MrCreosote:
Not clearing the Auth DB obviously, unnecessary and annoying

### DIFF
--- a/src/us/kbase/workspace/test/controllers/arango/ArangoController.java
+++ b/src/us/kbase/workspace/test/controllers/arango/ArangoController.java
@@ -14,6 +14,8 @@ import java.util.Scanner;
 import org.apache.commons.io.FileUtils;
 
 import com.arangodb.ArangoDB;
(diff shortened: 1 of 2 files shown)


sychan (approved):
Seems benign enough - for a bulk delete of the database. There's no way this could be accidentally pointed at a shared DB, right?

MrCreosote:
The client in the `ArangoController` is pointed at the arango instance that the controller starts up, so if someone changed the controller code so that the client pointed at another database and provided the creds for that database in the code, and that database was accessible from that someone's computer, then it could delete the database. 

If you have the creds and network access to the DB I can't see why you'd go to all that trouble though, just do it in `ipython` or the arango UI or something.

MrCreosote:
Here's how the client gets started for reference: https://github.com/MrCreosote/workspace_deluxe/blob/dev-candidate/src/us/kbase/workspace/test/controllers/arango/ArangoController.java#L49-L72

PR #389, opened 2020-10-15
Fix bug in sample ID handler

MrCreosote:
String interning caused `==` to work when the correct equality method is
using `equals()`

### DIFF
--- a/src/us/kbase/workspace/test/kbase/SampleIDHandlerFactoryTest.java
+++ b/src/us/kbase/workspace/test/kbase/SampleIDHandlerFactoryTest.java
@@ -331,7 +331,9 @@ public class SampleIDHandlerFactoryTest {
 		final SampleServiceClientWrapper cli = mock(SampleServiceClientWrapper.class);
 		
 		final IdReferenceHandler<Integer> h = new SampleIdHandlerFactory(cli)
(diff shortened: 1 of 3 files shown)


src/us/kbase/workspace/test/kbase/SampleIDHandlerFactoryTest.java
```diff
@@ -331,7 +331,9 @@ public void sampleIDHandlerProcessIDs() throws Exception {
 		final SampleServiceClientWrapper cli = mock(SampleServiceClientWrapper.class);
 		
 		final IdReferenceHandler<Integer> h = new SampleIdHandlerFactory(cli)
-				.createHandler(Integer.class, new AuthToken("t", "user1"));
+				// keep the new String(). String interning caused this test to pass when
+				// it should've failed
+				.createHandler(Integer.class, new AuthToken("t", new String("user1")));
```
sychan, line 336: Pretty subtle bug

sychan (approved):

src/us/kbase/workspace/test/kbase/SampleIDHandlerFactoryTest.java
MrCreosote: it's subtle here, but in the integration tests which are coming in the next PR, it caused authorization for the sample owner to fail

PR #391, opened 2020-10-15
Update Shock client to 0.1.0 for blobstore compatibility

MrCreosote:
As near as I can tell, how this screwup happened is:

* while developing the blobstore, I came up with a list of repos that required a shock client update.
* I left the workspace off that list since it would no longer talk to shock, but talk to S3 directly.
* Later, I added support for the `@bytestream` annotation that causes the workspace to talk to shock, but reasoned the client didn't need an update because the `@bytestream` code doesn't use any of the changed APIs.
* I forgot that on client startup the client *does* use one of the changed APIs invisibly to check the shock connection and credentials are fine.
* Apparently I never tested the bytestream stuff against the blobstore.

Lesson learned: always test everything against everything.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -22,9 +22,6 @@
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="lib" path="/jars/lib/jars/bson4jackson/bson4jackson-2.2.0-2.2.0.jar"/>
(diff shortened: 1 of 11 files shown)


kkellerlbl (approved):
LGTM

MrCreosote:
For posterity's sake, the update list actually *did* mention the bytestream stuff in its first iteration, but made the assumption that since the bytestream stuff didn't use any changed APIs it was fine. Famous last words.

PR #397, opened 2020-10-27
Add a trivial spec to make ad hoc testing easier

MrCreosote:
Frequently start up the services via the test with a debug pause and
then do manual testing. Having a simple spec makes this easier.


sychan (commented):
Do you want to comment on how the "don't use this" stuff can help with testing?

MrCreosote:
@sychan yep, that's confusing. Fixed

sychan (approved):

PR #407, opened 2021-01-07
Get rid of dev-candidate

MrCreosote:
Becomes a merge / deploy nightmare with features competing to be the
next one out. Better to have separate feature branches.


MrCreosote:
Docker build fail expected

PR #414, opened 2021-04-27
Remove an old TODO and fix a typo

MrCreosote:
Apparently forgot to remove the TODO when the integration tests were
finished


MrCreosote:
We expect the image build to fail on PRs - it should run successfully after the merge.

Issue #418, opened 2021-05-10
Remove dynamic service support for sample service


MrCreosote:
Is now a core service. Remove the release tag parameter and special code to switch between dynamic service and core service modes. Should be able to use a standard SDK compiled client, which simplifies the codebase a bit.

PR #419, opened 2021-07-13
Remove dynamic option for sample service

MrCreosote:
Sample service is now a core service and there are no plans to return it
to a dynamic service, so simplify the code and config by removing
dynamic service support.

Closes https://github.com/kbase/workspace_deluxe/issues/418


MrCreosote:
image build is expected to fail

scanon (approved):
LGTM

PR #421, opened 2021-07-15
Remove Shock as a workspace backend

MrCreosote:
Shock is still used to test the handle and bytestream annotation
integrations. Swapping that for the Blobstore is another, much more
annoying, PR.


MrCreosote:
image
It's... so beautiful

MrCreosote:
Image build is expected to fail

sychan (approved):
It looks sensible enough, let's see if it breaks anything!

sychan:
ps.
    You could have merged it into a branch in this repo and merged from there to see if the tokens work.

MrCreosote:
I'm not really worried about the tokens, they're just for building the image, which isn't needed pre-merge.

PR #422, opened 2021-07-21
Hide object notifications part 1: Remove search prototype event handler

MrCreosote:
Search now uses Kafka and the prototype was never intended to be a long
term solution anyway

Removing this makes it easier to add new events to the event listener
framework


MrCreosote:
Image build was expected to fail

ialarmedalien (approved):
👍

PR #424, opened 2021-07-23
Hide object notifications part 2: Add a hide event to the event listener framework

MrCreosote:
Nothing actually calls the new method yet, that'll happen in future PRs


ialarmedalien (approved):
✨

PR #428, opened 2021-07-27
Return object modification times on a hide call

MrCreosote:
Needed for event handling, which requires a time.

### DIFF
--- a/src/us/kbase/workspace/test/database/mongo/MongoWorkspaceDBTest.java
+++ b/src/us/kbase/workspace/test/database/mongo/MongoWorkspaceDBTest.java
@@ -243,26 +243,10 @@ public class MongoWorkspaceDBTest {
 				inst(90000)  // update other ws date
 				);
 		
-		// create workspaces
-		final WorkspaceUser u = new WorkspaceUser("u");
-		db.createWorkspace(u, "ws1", false, null, new WorkspaceUserMetadata());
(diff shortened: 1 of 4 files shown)


src/us/kbase/workspace/test/database/mongo/MongoWorkspaceDBTest.java
```diff
@@ -345,6 +329,104 @@ private void checkWorkspaceUpdateTimes(final Set<Instant> wsTimes) {
...
+		final ResolvedWorkspaceID wsid = new ResolvedWorkspaceID(1, "ws", false, false);
+		final ResolvedWorkspaceID wsid2 = new ResolvedWorkspaceID(2, "ws2", false, false);
+
+		final Set<ObjectIDResolvedWS> objectIDs = setupTestDataForHideDelete(db, wsid, wsid2);
+
+		hideOrUnhideAndCheck(db, objectIDs, true, wsid, wsid2, set(inst(20000), inst(40000)));
+		hideOrUnhideAndCheck(db, objectIDs, false, wsid, wsid2, set(inst(60000), inst(80000)));
+	}
+	
+	private void hideOrUnhideAndCheck(
```
MrCreosote, line 387: I tried combining this method and the equivalent method for delete, but it became an unreadable mess with a zillion method arguments (and there's already a lot)

ialarmedalien (approved):
👍

PR #429, opened 2021-07-27
Fire hide events

MrCreosote:


ialarmedalien (approved):
Looks more than reasonable!

PR #434, opened 2021-09-27
Remove unused cruft, force cursor close

MrCreosote:
For some reason getting the list of workspaces to which a user had
access allows specifying multiple users, which confused the hell out of
me. Turns out this feature and a couple of helper methods are unused, so
in the garbage they go.

Also some other minor cruft cleanup and force cursors to close.

### DIFF
--- a/src/us/kbase/typedobj/db/MongoTypeStorage.java
+++ b/src/us/kbase/typedobj/db/MongoTypeStorage.java
@@ -375,10 +375,11 @@ public class MongoTypeStorage implements TypeStorage {
 			final DBObject whereCondition,
 			final BasicDBObject projection,
 			final Class<T> clazz) {
-		final DBCursor find = col.find(whereCondition, projection);
 		final List<T> data = new LinkedList<>();
-		for (final DBObject dbo: find) {
-			data.add(toObj(dbo, clazz));
+		try (final DBCursor find = col.find(whereCondition, projection)) {
+			for (final DBObject dbo: find) {
+				data.add(toObj(dbo, clazz));
+			}
 		}
 		return data;
 	}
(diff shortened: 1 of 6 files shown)


src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
```diff
@@ -1070,17 +1070,9 @@ public ResolvedWorkspaceID resolveWorkspace(
...
-			throws WorkspaceCommunicationException,
-			CorruptWorkspaceDBException {
-		final Map<ResolvedWorkspaceID, Map<User, Permission>> res =
-				query.queryPermissions(rwsis, null);
-		final Map<ResolvedWorkspaceID, Map<User, Permission>> ret = 
-				new HashMap<ResolvedWorkspaceID, Map<User, Permission>>();
-		//probably a better way to do this
-		for (final ResolvedWorkspaceID r: res.keySet()) {
-			ret.put((ResolvedWorkspaceID)r, res.get(r)); 
-		}
```
MrCreosote, line 1082: This is bizarre - lines 1077 on are a no-op. I think this is cruft from a completely crazy initial design that subclassed the resolved WS ID somehow, but I forget the details.

src/us/kbase/typedobj/db/MongoTypeStorage.java
```diff
@@ -375,10 +375,11 @@ public String getModuleSpecRecord(String moduleName, long version)
 			final DBObject whereCondition,
 			final BasicDBObject projection,
 			final Class<T> clazz) {
-		final DBCursor find = col.find(whereCondition, projection);
 		final List<T> data = new LinkedList<>();
-		for (final DBObject dbo: find) {
-			data.add(toObj(dbo, clazz));
+		try (final DBCursor find = col.find(whereCondition, projection)) {
```
MrCreosote, line 379: This `try()` format is the equivalent of the `with` statement in python. In this case it ensures that the `DBCursor` is always closed when the block exists. There are several of these changes in this PR.

sychan (approved):

PR #435, opened 2021-09-27
Clarify handle error

MrCreosote:
Confusing when the handle is readable but you can't save an object
containing the handle ID.


Tianhao-Gu (approved):
👍 Much cleaner error message.

MrCreosote:
Considering this is a text only change, the current codecov comment doesn't make much sense

PR #437, opened 2021-09-30
Required workspaces in list_objects, refactor parameters

MrCreosote:
This PR turned out to be really big. Sorry about that.

It does two major things, which wound up having a number of knock on effects:

1. It refactors the list object parameters class to use a builder rather than a zillion constructors.
    1. It also modernized the class a bit, using Optional and Instant instead of nulls and Dates.
2. It requires between 1 and 10000 workspaces as inputs when listing objects. Type only lists are dangerous because a large number of workspaces (potentially >> 10000) may need to be examined as part of the operation.

It didn't really make sense to not combine these two things, since refactoring the class means redoing the constructors, which is what the workspace requirements would do.

That has some knock on effects:
1. the minimum permission parameter for list objects is now useless, since a workspace list is required, so the user should just filter there.
2. likewise, the exclude global workspaces command is also useless.
3. Some other more minor changes in the part of the codebase  surrounding the parameters class.


src/us/kbase/workspace/test/workspace/WorkspaceTest.java
```diff
@@ -605,68 +605,27 @@ public void adminListObjects() throws Exception {
...
 
-		lop = new ListObjectsParameters(Arrays.asList(wsi, wsi2), SAFE_TYPE1)
-				.withIncludeMetaData(true).withShowDeleted(true);
+		lop = ListObjectsParameters.getBuilder(Arrays.asList(wsi, wsi2)).withType(SAFE_TYPE1)
+				.withIncludeMetaData(true).withShowDeleted(true).withAsAdmin(true).build();
 		compareObjectInfo(ws.listObjects(lop), Arrays.asList(del, std2));
 	}
 	
-	@Test
-	public void adminListObjectsFailConstructParams() {
```
MrCreosote, line 629: Made redundant by the new unit tests for the class

src/us/kbase/workspace/test/workspace/WorkspaceTest.java
```diff
@@ -5890,64 +5851,32 @@ public void listObjectsWithDeletedObjects() throws Exception {
...
-		compareObjectInfo(ws.listObjects(lop.withShowOnlyDeleted(false)), Arrays.asList(std));
-		compareObjectInfo(ws.listObjects(lop.withShowDeleted(true)),
-				Arrays.asList(std, del));
-		compareObjectInfo(ws.listObjects(lop.withShowDeleted(false).withShowOnlyDeleted(true)),
-				Arrays.asList(del));
-		
-	}
-	
-	@Test
-	public void listObjectsWithDeletedWorkspace() throws Exception {
```
MrCreosote, line 5922: This scenario can no longer occur

src/us/kbase/workspace/test/workspace/WorkspaceTest.java
```diff
@@ -6076,224 +6009,185 @@ public void listObjectsAndHistory() throws Exception {
...
 				Arrays.asList(stdws2, hidden, deleted));
-		compareObjectInfo(ws.listObjects(new ListObjectsParameters(user2, allType2)
-				.withShowHidden(true).withShowDeleted(true)
-				.withShowAllVersions(true).withIncludeMetaData(true)),
+		compareObjectInfo(ws.listObjects(ListObjectsParameters.getBuilder(allws2)
+				.withUser(user2).withType(allType2).withShowHidden(true).withShowDeleted(true)
+				.withShowAllVersions(true).withIncludeMetaData(true).build()),
 				new ArrayList<ObjectInformation>());
 		
-		//TODO MOVE move these to unit tests for LOP
```
MrCreosote, line 6264: Made redundant by the new unit tests for the class

MrCreosote:
Travis didn't pick up the PR it seems; closing & reopening fixed it

lib/Bio/KBase/workspace/Client.pm
```diff
@@ -10921,21 +10921,21 @@ auth has a value which is a string
 
 Parameters for the 'list_objects' function.
 
-                At least one of the following filters must be provided. It is strongly
-                recommended that the list is restricted to the workspaces of interest,
-                or the results may be very large:
+                At least one, and no more than 10000, workspaces musts be specified in one of the
```
sychan, line 10924: "musts"? Is Gollum writing your docs?

docsource/releasenotes.rst
```diff
@@ -6,6 +6,10 @@ VERSION: 0.12.0 (Released TBD)
 
 BACKWARDS INCOMPATIBILITIES:
 
+* ``list_objects`` now requires that at least one and no more than 10000 workspaces are
+  specified.
+* As such, the ``excludeGlobal`` and ``perm`` parameters no longer have any use and are
```
sychan, line 11: Are they just being silently ignored? I didn't find anything in the PR that seems to generates errors or warnings for the deprecated params.

lib/Bio/KBase/workspace/Client.pm
sychan: Also, what happens if someone puts in 10001 workspaces? Is the 10k limit enforced on the object type, or does the method throw an error?

MrCreosote: > "musts"? Is Gollum writing your docs?
I actually had this fixed in the spec but forgot to recompile. Anyway, fixed

docsource/releasenotes.rst
MrCreosote: Yes, silently ignored. I don't want to remove them and have code start failing unnecessarily. There's no way to return warnings to the user without backwards incompatible changes to the API, unfortunately. I could log the usage, but I'm not sure what we'd do with those logs

PR #438, opened 2021-10-04
Fix test name

MrCreosote:
Match to class being tested


MrCreosote:
Approved out of band by @sychan

PR #441, opened 2021-10-06
Move resolved list objects params into main class & test

MrCreosote:
Makes more sense to keep class B, which can only be created by class A,
internal to class A.

Also adds unit tests.


src/us/kbase/workspace/database/ListObjectsParameters.java
```diff
@@ -191,16 +191,179 @@ public boolean asAdmin() {
...
 				minObjectID, maxObjectID, showHidden, showDeleted, showOnlyDeleted, showAllVers,
 				includeMetaData, limit, asAdmin);
 		
 	}
 	
+	/** Parameters for listing objects resolved with workspace permissions for the user.
+	 * @author crushingismybusiness
+	 *
+	 */
+	public static class ResolvedListObjectParameters {
```
MrCreosote, line 206: Basically identical to the deleted GetObjectInformationParameters class above.

PR #445, opened 2021-10-08
Add method to get type string w/ major version only

MrCreosote:
Will be used for setting up new fields for indexes for paging on types


MrCreosote:
The type def classes really need cleanup, but it'd be hours of work and I've already done a lot of cleanup on this project so... later

PR #448, opened 2021-10-12
Add type filtered sorting to list objects + unit tests 

MrCreosote:
Adds the ability to filter on a type and sort based on the object UPA,
which in turn will allow paging in a future PR.

Also adds unit tests for the ObjectLister class, which it really needs
because if a sort specification gets applied to a query that needs a
index other than the one that supplies the sort to perform well it could
really slow things down

Will add integration tests to confirm the sort is working as expected in
the next PR as this one is already pretty large.

### DIFF
--- a/src/us/kbase/workspace/database/mongo/ObjectLister.java
+++ b/src/us/kbase/workspace/database/mongo/ObjectLister.java
@@ -9,7 +9,9 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -135,17 +137,7 @@ public class ObjectLister {
 	private DBObject buildSortSpec(final ResolvedListObjectParameters params) {
 		final DBObject sort = new BasicDBObject();
 		if (isSafeForUPASort(params)) {
-			if (params.getType().isPresent()) {
-				final TypeDefId t = params.getType().get();
-				// TODO CODE use optionals for typedefid versions
-				if (t.getMinorVersion() != null) {
-					sort.put(Fields.VER_TYPE_FULL, 1);
-				} else if (t.getMajorVersion() != null) {
-					sort.put(Fields.VER_TYPE_WITH_MAJOR_VERSION, 1);
-				} else {
-					sort.put(Fields.VER_TYPE_NAME, 1);
-				}
-			}
+			addTypeField(sort, params.getType(), () -> 1);
 			sort.put(Fields.VER_WS_ID, 1);
 			sort.put(Fields.VER_ID, 1);
 			sort.put(Fields.VER_VER, -1);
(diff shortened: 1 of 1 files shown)


src/us/kbase/workspace/database/mongo/ObjectInfoUtils.java
```diff
@@ -38,15 +38,15 @@
...
+	public ObjectInfoUtils(final QueryMethods query) {
 		this.query = requireNonNull(query, "query argument may not be null");
 	}
 	
 	private static final Set<String> FLDS_LIST_OBJ = Stream.of(
 			Fields.OBJ_ID, Fields.OBJ_NAME, Fields.OBJ_DEL, Fields.OBJ_HIDE,
 			Fields.OBJ_VCNT, Fields.OBJ_WS_ID).collect(Collectors.toSet());
 	
-	Map<Map<String, Object>, ObjectInformation> generateObjectInfo(
+	public Map<Map<String, Object>, ObjectInformation> generateObjectInfo(
```
MrCreosote, line 49: Allow mocking

src/us/kbase/workspace/database/mongo/ObjectLister.java
```diff
@@ -65,12 +81,11 @@ public ObjectLister(final DBCollection verCol, final ObjectInfoUtils infoUtils)
...
 		//condition where the workspace object was saved but no versions
 		//were saved yet
-		
-		final List<ObjectInformation> ret = new LinkedList<>();
 		try (final DBCursor cur = verCol.find(verq, projection)) {
 			cur.sort(sort);
+			final List<Map<String, Object>> verobjs = new ArrayList<>(querysize);
 			while (cur.hasNext() && ret.size() < params.getLimit()) {
-				final List<Map<String, Object>> verobjs = new ArrayList<>();
+				verobjs.clear();
```
MrCreosote, line 88: Minor memory usage improvement, reuse allocated array vs. allocating a new array every time

src/us/kbase/workspace/database/mongo/ObjectLister.java
```diff
@@ -119,39 +134,52 @@ private DBObject buildProjection(final ResolvedListObjectParameters params) {
...
-				&& !params.getType().isPresent();
-		return oidFiltersOnly;
+				&& params.getSavers().isEmpty();
 	}
 
 	private DBObject buildQuery(final ResolvedListObjectParameters params) {
-		final Set<Long> ids = params.getPermissionSet().getWorkspaces().stream()
-				.map(ws -> ws.getID()).collect(Collectors.toSet());
+		final List<Long> ids = params.getPermissionSet().getWorkspaces().stream()
+				.map(ws -> ws.getID()).distinct().sorted().collect(Collectors.toList());
```
MrCreosote, line 171: Since the query is passed to the ObjectInfoUtils mock, my choices are 1) write a custom matcher which I know from experience is verbose, annoying, and hard to read, or 2) sort here so the resulting list is deterministic. Since MongoDB will sort `$in` lists anyway, this moves a trivial amount of load from the db to this application for the cost of a trivial amount of memory, so win / win, if not very large wins.

The workspaces in the permission set *should* be unique, but looking at the code that's not guaranteed, hence the `distinct()` call.

src/us/kbase/workspace/database/mongo/ObjectLister.java
```diff
@@ -119,39 +134,52 @@ private DBObject buildProjection(final ResolvedListObjectParameters params) {
...
-			verq.put(Fields.VER_TYPE_FULL, new BasicDBObject(
-					"$regex", "^" + params.getType().get().getTypePrefix()));
+			final TypeDefId t = params.getType().get();
+			if (t.getMinorVersion() != null) {
+				verq.put(Fields.VER_TYPE_FULL, t.getTypeString());
+			} else if (t.getMajorVersion() != null) {
+				verq.put(Fields.VER_TYPE_WITH_MAJOR_VERSION, t.getTypeString());
+			} else {
+				verq.put(Fields.VER_TYPE_NAME, t.getTypeString());
+			}
```
ialarmedalien, line 182: since this is very similar to the code around line 137, is it worth abstracting it out into a function?

src/us/kbase/workspace/test/database/mongo/ObjectListerTest.java
```diff
@@ -0,0 +1,796 @@
...
+		when(cur.next()).thenReturn(OBJ_DB_1);
+		
+		when(m.infoutils.generateObjectInfo(
+				pset,
+				Arrays.asList(OBJ_MAP_1),
+				boolopts.get(1),
+				boolopts.get(2),
+				boolopts.get(3),
+				boolopts.get(4),
+				boolopts.get(5)))
```
ialarmedalien, line 290: java doesn't have any nicer way to do this kind of sequence?

src/us/kbase/workspace/database/mongo/ObjectLister.java
MrCreosote: Done

PR #449, opened 2021-10-13
Extract ServerThread test class for reuse

MrCreosote:
Thought I was going to need this for the tests I'm currently writing, but it turns out I don't, but I like already did the work so like idk whatevs


ialarmedalien (approved):
I'm going to make the rash assumption that you have thoroughly checked the codebase and found every instance that requires replacement. 👍

MrCreosote:
I didn't have to, Eclipse marked all the locations for me.

PR #450, opened 2021-10-13
Remove auth server from WorkspaceTester test rig

MrCreosote:
Was only needed for Shock, which is no longer a supported backend.

Another discovery I made while figuring out how to set up a new integration test class, which I haven't done for the workspace in years


ialarmedalien (approved):
👍

PR #455, opened 2021-10-20
Add string parsing method to create a RefLimit

MrCreosote:


src/us/kbase/workspace/database/RefLimit.java
```diff
@@ -50,6 +50,53 @@ public static RefLimit build(final Long wsid, final Long objid, final Integer ve
...
+		}
+		return new RefLimit(ws, id, ver);
+	}
+
+	private static Long parseLong(
+			final String putativeNum,
+			final String refstring,
+			final String name) {
+		try {
+			return Long.parseLong(putativeNum);
```
ialarmedalien, line 93: What happens with decimals?

MrCreosote: It'll cause an error. I added some test cases to cover that.

PR #457, opened 2021-10-20
Revamp indexing strategy for types

MrCreosote:
The old strategy required duplicating the type name 3 times in the
object version document. The new strategy has 2 duplications and longer
term one of them can be removed.

### DIFF
--- a/src/us/kbase/workspace/database/mongo/Fields.java
+++ b/src/us/kbase/workspace/database/mongo/Fields.java
@@ -55,11 +55,14 @@ public class Fields {
 	public static final String VER_REF = "refs";
 	public static final String VER_PROVREF = "provrefs";
 	// the full type string, e.g. Module.Type-3.4
+	// TODO CODE kept to allow for rollbacks, but unused at this point. remove in future
 	public static final String VER_TYPE_FULL = "type";
-	// the type string with the major version, e.g. Module.Type-3
-	public static final String VER_TYPE_WITH_MAJOR_VERSION = "tymaj";
 	// the type name only as a type string, e.g. Module.Type
 	public static final String VER_TYPE_NAME = "tyname";
+	// the major version of the type, e.g. 3
+	public static final String VER_TYPE_MAJOR_VERSION = "tymaj";
+	// the minor version of the type, e.g. 4
+	public static final String VER_TYPE_MINOR_VERSION = "tymin";
 	public static final String VER_SIZE = "size";
 	public static final String VER_RVRT = "revert";
 	public static final String VER_META = "meta";
--- a/src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
+++ b/src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
@@ -41,6 +41,7 @@ import us.kbase.typedobj.core.ExtractedMetadata;
 import us.kbase.typedobj.core.MD5;
 import us.kbase.typedobj.core.SubsetSelection;
 import us.kbase.typedobj.core.TempFilesManager;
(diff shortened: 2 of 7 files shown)


src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
```diff
@@ -1581,14 +1622,6 @@ private void saveObjectVersions(final WorkspaceUser user,
 			v.put(Fields.VER_WS_ID, wsid.getID());
 			v.put(Fields.VER_ID, objectid);
 			v.put(Fields.VER_VER, ver++);
-			if (!v.containsKey(Fields.VER_TYPE_NAME)) {
```
MrCreosote, line 1584: Originally my idea was to allow an online schema update post server upgrade, but then I realized that would mean any `list_objects` calls with a type filter would have missing data until the update was complete. As such, the server will be modified not to start until the update is complete, and this code isn't necessary.

src/us/kbase/workspace/database/mongo/Fields.java
```diff
@@ -55,11 +55,14 @@
...
+	// TODO CODE kept to allow for rollbacks, but unused at this point. remove in future
 	public static final String VER_TYPE_FULL = "type";
-	// the type string with the major version, e.g. Module.Type-3
-	public static final String VER_TYPE_WITH_MAJOR_VERSION = "tymaj";
 	// the type name only as a type string, e.g. Module.Type
 	public static final String VER_TYPE_NAME = "tyname";
+	// the major version of the type, e.g. 3
+	public static final String VER_TYPE_MAJOR_VERSION = "tymaj";
+	// the minor version of the type, e.g. 4
+	public static final String VER_TYPE_MINOR_VERSION = "tymin";
```
ialarmedalien, line 65: strangely enough, I was planning to name the first products of my horrific biohacking project Tymin and Tymaj.

src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
```diff
@@ -331,6 +333,42 @@ private static IndexSpecification idxSpec(
 				getIndexOptions(options));
 	}
 	
+	// ew. Not sure how to make this less gross though.
```
ialarmedalien, line 336: I can think of options for other languages but don't think they would translate to Java 😕

src/us/kbase/workspace/test/database/mongo/ObjectListerTest.java
```diff
@@ -144,7 +148,8 @@ private DBObject getProjection() {
 	
 	private DBObject getProjection(final boolean includeMeta) {
 		final BasicDBObject p = new BasicDBObject();
-		Stream.of("ver", "type", "savedate", "savedby", "chksum", "size", "id", "ws")
+		Stream.of("ver", "tyname", "tymaj", "tymin", "savedate", "savedby", "chksum", "size",
+				"id", "ws")
```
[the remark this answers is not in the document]
ialarmedalien: Er, what I meant to say was create a set of variables representing those strings. Variable-ise?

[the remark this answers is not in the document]
MrCreosote: For actual code I would agree but for tests I'm less concerned about it since the tests should fail if there's a typo, and making the same typo in many places (and thus allowing for the possibility that all tests could pass when they shouldn't) seems improbable. 

In general I tend to shy away from too much variableization and DRYing in tests because there's a tendency to wind up recapitulating  the actual code in the test code. Hence I try to use literals and construct data structures per test and avoid helper methods that, again, start recapitulating the code under test.

PR #460, opened 2021-10-27
Switch from the Mongo 2.X api to the 3.0 API.

MrCreosote:
Not really any way to do the switch piecemeal - once you switch from
`getDB()` to `getDatabase()` you're on the new API and have to make changes
everywhere.

Lots and lots of potential refactoring was unearthed while making these
changes, but was only done where not doing it would have been more work
in order to minimize the changed line count, which is already large.

### DIFF
--- a/src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
+++ b/src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
@@ -32,6 +32,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.Set;
 
+import org.bson.Document;
 import org.bson.types.ObjectId;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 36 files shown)


src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
```diff
@@ -461,17 +489,17 @@ private static void checkConfig(final DB wsdb) throws WorkspaceCommunicationExce
...
 	}
 	
-	static void ensureIndexes(final DB wsdb) throws CorruptWorkspaceDBException {
+	static void ensureIndexes(final MongoDatabase wsdb) throws CorruptWorkspaceDBException {
 		final HashMap<String, List<IndexSpecification>> indexes = getIndexSpecs();
 		for (final String col: indexes.keySet()) {
 			for (final IndexSpecification index: indexes.get(col)) {
 				try {
 					wsdb.getCollection(col).createIndex(index.index, index.options);
+					// this seems to be the 1 *&)&*)* place where DKE is still used
```
ialarmedalien, line 502: hmmm, decoding `*&)&*)*` could be tricky for other developers... perhaps you should leave a hint as to how they can crack the code?

ialarmedalien (approved):
Apart from the stray space, I couldn't see any obvious issues.

I'll send you the therapy bills.

MrCreosote:
I've already got a stack of commits on top of this one, so I've fixed the two issues you brought up on the top of the stack. You'll see them in a few PRs.

PR #462, opened 2021-10-29
Move WS DB dependency to class where it's actually used

MrCreosote:
This was too stupid not to fix. Maybe `MongoWorkspaceDB` at some point
used `TempFilesManager` but now it doesn't touch TFM. Make the
`Workspace` class responsible for it since WS actually uses it.

### DIFF
--- a/src/us/kbase/workspace/test/workspace/WorkspaceListenerTest.java
+++ b/src/us/kbase/workspace/test/workspace/WorkspaceListenerTest.java
@@ -26,6 +26,7 @@ import org.mockito.ArgumentMatcher;
 import com.google.common.collect.ImmutableMap;
 
 import us.kbase.common.service.UObject;
+import us.kbase.typedobj.core.TempFilesManager;
 import us.kbase.typedobj.core.TypeDefId;
 import us.kbase.typedobj.core.TypedObjectValidator;
 import us.kbase.typedobj.core.ValidatedTypedObject;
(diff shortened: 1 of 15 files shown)


src/us/kbase/workspace/test/workspace/WorkspaceListenerTest.java
```diff
@@ -26,6 +26,7 @@
 import com.google.common.collect.ImmutableMap;
```
MrCreosote, line 26: Lots of cleanup possible for this file, but the current cleanup helps a lot and is 400 lines so... later

src/us/kbase/workspace/test/workspace/WorkspaceUnitTest.java
```diff
@@ -65,16 +67,20 @@ private TestMocks initMocks() {
 		@SuppressWarnings("unused")
 		private final ResourceUsageConfiguration cfg;
 		private final Workspace ws;
+		@SuppressWarnings("unused")
+		private TempFilesManager tfm;
```
MrCreosote, line 71: this actually is used to create the workspace, so I'd prefer to include it in the TestMocks class so the next person doesn't have to dig around if they need to use the tfm mock

ialarmedalien (approved):
I'm assuming with PRs like this, you can find all occurrences that need to be replaced using your IDE. LGTM. 👍

MrCreosote:
With a change like this the code straight up won't compile unless I fix it everywhere

PR #464, opened 2021-11-01
Extract test object save helper method for reuse

MrCreosote:
Designed for use with the PartialMock set up so added to that class


ialarmedalien (approved):
Looks uncontroversial. I expect better next time!

PR #468, opened 2021-11-02
Support loading config from a file

MrCreosote:
The server super class basically takes care of this for the workspace
server, but that doesn't help if you want to load the config without the
server running

### DIFF
--- a/src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
+++ b/src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
@@ -198,6 +198,7 @@ public class KBaseWorkspaceConfig {
 		}
 	}
 	
+	/** An error encountered when attempting to create a workspace configuration. */
 	@SuppressWarnings("serial")
 	public static class KBaseWorkspaceConfigException extends Exception {
 		
@@ -217,6 +218,11 @@ public class KBaseWorkspaceConfig {
 		}
 	}
 	
+	/** Create the configuration from an *.ini file.
+	 * @param configFile the file.
+	 * @param configSection the section of the file containing the configuration.
+	 * @throws KBaseWorkspaceConfigException if creating the configuration fails.
+	 */
 	public KBaseWorkspaceConfig(final Path configFile, final String configSection)
 			throws KBaseWorkspaceConfigException {
 		this(getConfigMap(configFile, configSection));
--- a/src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
+++ b/src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
@@ -3,8 +3,10 @@ package us.kbase.workspace.kbase;
 import static java.util.Objects.requireNonNull;
 import static us.kbase.workspace.database.Util.checkString;
 
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -16,9 +18,12 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.ini4j.Ini;
+
 import com.google.common.collect.ImmutableMap;
 
 import software.amazon.awssdk.regions.Region;
+import us.kbase.common.service.JsonServerServlet;
 
 public class KBaseWorkspaceConfig {
 	
@@ -192,7 +197,56 @@ public class KBaseWorkspaceConfig {
 			return true;
 		}
 	}
+	
+	@SuppressWarnings("serial")
+	public static class KBaseWorkspaceConfigException extends Exception {
+		
+		/** Create the exception.
+		 * @param message the exception message.
+		 */
+		public KBaseWorkspaceConfigException(final String message) {
+			super(message);
+		}
+		
+		/** Create the exception with a cause.
+		 * @param message the exception message.
+		 * @param cause the cause of this exception.
+		 */
+		public KBaseWorkspaceConfigException(final String message, final Throwable cause) {
+			super(message, cause);
+		}
+	}
+	
+	public KBaseWorkspaceConfig(final Path configFile, final String configSection)
+			throws KBaseWorkspaceConfigException {
+		this(getConfigMap(configFile, configSection));
+	}
 
+	private static Map<String, String> getConfigMap(
+			final Path configFile,
+			final String configSection)
+			throws KBaseWorkspaceConfigException {
+		final Ini ini;
+		try {
+			ini = new Ini(configFile.toFile());
+		} catch (IOException ioe) {
+			throw new KBaseWorkspaceConfigException(
+					String.format("Could not read from configuration file %s: %s",
+							configFile, ioe.getLocalizedMessage()),
+					ioe);
+		}
+		final Map<String, String> config = ini.get(configSection);
+		if (config == null) {
+			throw new KBaseWorkspaceConfigException(
+					String.format("The configuration file %s has no section %s",
+							configFile, configSection));
+		}
+		return config;
+	}
+
+	/** Create the configuration from a mapping as provided by {@link JsonServerServlet}.
+	 * @param config the configuration
+	 */
 	public KBaseWorkspaceConfig(final Map<String, String> config) {
 		// this is a really long constructor. Clean up at some point
 		requireNonNull(config, "config");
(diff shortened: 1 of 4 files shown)


src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
```diff
@@ -89,6 +122,138 @@ private void listenerConstructFail(
...
+		} catch (Exception got) {
+			TestCommon.assertExceptionCorrect(got, expected);
+		}
+	}
+	
+	// TODO NOW_NEXT_PR reorder the args to be more sensible
+	// good heavens, what what absolutely ghastly code. I may faint
+	// hey kids, don't try this at home
+	private void assertConfigCorrect(
+			final Map<String, String> cfg,
```
MrCreosote, line 171: These arguments are in the order of the original asserts in the tests (other than one case)

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
```diff
@@ -110,39 +275,39 @@ public void configMinimal() throws Exception {
...
-		assertThat("incorrect bytestream user", kwc.getBytestreamUser(), nullValue());
-		assertThat("incorrect sample token", kwc.getSampleServiceToken(), nullValue());
-		assertThat("incorrect sample url", kwc.getSampleServiceURL(), nullValue());
-		assertThat("incorrect temp dir", kwc.getTempDir(), is("temp"));
-		assertThat("incorrect type db", kwc.getTypeDBName(), is("typedb"));
-		assertThat("incorrect ws admin", kwc.getWorkspaceAdmin(), nullValue());
-		assertThat("incorrect has err", kwc.hasErrors(), is(false));
-		assertThat("incorrect ignore hs", kwc.ignoreHandleService(), is(true));
+		assertConfigCorrect(
+				cfg,									// config under test
```
MrCreosote, line 279: The comments line up beautifully with a 4 space tab

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
```diff
@@ -433,41 +581,41 @@ public void configFailNullRequiredEntries() throws Exception {
...
-		assertThat("incorrect bytestream user", kwc.getBytestreamUser(), nullValue());
-		assertThat("incorrect sample token", kwc.getSampleServiceToken(), nullValue());
-		assertThat("incorrect sample url", kwc.getSampleServiceURL(),
-				is(new URL("https://foo.com")));
-		assertThat("incorrect temp dir", kwc.getTempDir(), nullValue());
-		assertThat("incorrect type db", kwc.getTypeDBName(), nullValue());
-		assertThat("incorrect ws admin", kwc.getWorkspaceAdmin(), nullValue());
-		assertThat("incorrect has err", kwc.hasErrors(), is(true));
-		assertThat("incorrect ignore hs", kwc.ignoreHandleService(), is(false));
+		assertConfigCorrect(
```
MrCreosote, line 586: Watch out for this one, the asserts aren't in quite the same order as the rest of the tests

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
MrCreosote: Who uses an 8 space tab in 2021 anyway?

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
```diff
@@ -166,7 +468,7 @@ public void configMaximalS3() throws Exception {
 				.with("backend-token", "    token token token    ")
 				.with("backend-container", "   mahbukkit   ")
 				.with("backend-region", "   a-lovely-region   ")
-				.with("backend-trust-all-ssl-certificates", "trudat")
+				.with("backend-trust-all-ssl-certificates", "     true    ")
```
MrCreosote, line 471: So it turns out that a true case was never tested for this config item. Whoops. Shows the value of compiler warnings I guess

src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
```diff
@@ -583,4 +643,198 @@ public String getParamReport() {
...
+		if (obj == null)
+			return false;
+		if (getClass() != obj.getClass())
+			return false;
+		KBaseWorkspaceConfig other = (KBaseWorkspaceConfig) obj;
+		if (adminReadOnlyRoles == null) {
+			if (other.adminReadOnlyRoles != null)
+				return false;
+		} else if (!adminReadOnlyRoles.equals(other.adminReadOnlyRoles))
+			return false;
```
ialarmedalien, line 698: I don't know if there is a way to access these values through `this` -- as it is, you have a lot of repeated stanzas in the form
```
if (variable == null) {
    if (other.variable != null)
        return false;
} else if (!variable.equals(other.variable))
        return false;
}
```
and it would make life easier to be able to iterate through a list of properties, running that check for each of them.

This could also be turned into a static method that compares two objects -- that would make it much easier to do
```
// pseudocode
for (variable in variable_list) {
    if (obj1[variable] == null) {
        if (obj2[variable] != null)
            return false;
    } else if (obj1[variable] != obj2[variable]) {
        return false;
    }
}
```

Unfortunately I am not familiar enough with Java to know if magic of the type I have described here is possible...

src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
```diff
@@ -583,4 +643,198 @@ public String getParamReport() {
...
 	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ((adminReadOnlyRoles == null) ? 0 : adminReadOnlyRoles.hashCode());
+		result = prime * result + ((adminRoles == null) ? 0 : adminRoles.hashCode());
+		result = prime * result + ((auth2URL == null) ? 0 : auth2URL.hashCode());
+		result = prime * result + ((authURL == null) ? 0 : authURL.hashCode());
```
ialarmedalien, line 654: this seems like another function with a lot of similar declarations that could be compacted down to a list to be iterated through.

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
```diff
@@ -89,6 +121,295 @@ private void listenerConstructFail(
...
+			return this;
+		}
+
+		public ExpectedConfig withAdminRoles(final Set<String> adminRoles) {
+			this.adminRoles = adminRoles;
+			return this;
+		}
+
+		public ExpectedConfig withAuth2URL(final URL auth2url) {
+			auth2URL = auth2url;
```
ialarmedalien, line 211: no `this`?

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
```diff
@@ -89,6 +121,295 @@ private void listenerConstructFail(
...
+
+		public ExpectedConfig withAuthURL(final URL authURL) {
+			this.authURL = authURL;
+			return this;
+		}
+
+		public ExpectedConfig withBackendToken(final String backendToken) {
+			this.backendToken = backendToken;
+			return this;
+		}
```
ialarmedalien, line 223: do all these declarations have to be written out in full? There's no shortcut where you write a function to generate a whole load of similar properties?

src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
MrCreosote: `equals()` and `hashCode()` are both autogenerated other than the changes I made to get `Region` to work - any changes I made would be blown away the next time I made changes to the state variable and regenerated the functions, so I'd prefer to leave them as is.

src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
MrCreosote: See notes for equals() above

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
MrCreosote: The variable names are different so `this` isn't required. That being said, all these methods were autogenerated (sort of) and I didn't notice the variable name difference and I don't like it, so I've fixed it

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
MrCreosote: Without using https://projectlombok.org/, which I haven't learned yet but is on my list, not without doing some really crazy stuff with the classloader AFAIK.

src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
ialarmedalien: Sounds good!

ialarmedalien (approved):
Looks good!

PR #475, opened 2021-11-04
Add version decrementation to RefLimit

MrCreosote:
Step 1 of doing the stupid ref math serverside instead of making the
user do it


ialarmedalien (approved):
LGTM

Issue #486, opened 2021-11-18
Add option to skip external ACL updates to get_objects2


MrCreosote:
handle / blobstore / sample etc. updates can *really* slow down the data fetching, and aren't always necessary, especially when debugging or using the `no_data` option.

MrCreosote:
This is also useful for the case where an administrator is pulling data from the workspace and doesn't want to add themselves to the external ACLs, e.g. the search indexer

PR #490, opened 2021-11-23
Add param to skip external ACL updates

MrCreosote:
External ACL updates can take a significant portion of the data fetch time, especially when `no_data=1`.

Also useful for admins that don't want to add themselves to the ACL lists.

Fixes https://github.com/kbase/workspace_deluxe/issues/486


MrCreosote:
Closed / reopened to get Travis to run

src/us/kbase/workspace/test/kbase/JSONRPCLayerTest.java
```diff
@@ -904,7 +904,9 @@ public void saveAndGetObjects() throws Exception {
 				.withName("auto1");
 		Map<String, List<String>> exp = new HashMap<String, List<String>>();
 		ObjectData objo = CLIENT1.getObjects2(new GetObjects2Params()
-			.withObjects(toObjSpec(Arrays.asList(ojbid)))).getData().get(0);
+			.withObjects(toObjSpec(Arrays.asList(ojbid)))
+			.withSkipExternalAclUpdates(1L) // should have no effect
+			).getData().get(0);
```
ialarmedalien, line 909: is there another test where you verify that adding the skip external ACL updates param has no effect, either implicitly or explicitly?

src/us/kbase/workspace/kbase/ArgUtils.java
```diff
@@ -395,21 +393,34 @@ private static String formatDate(final Instant date) {
...
 			}
 			final PermError error = makeExternalIDsReadable(o, permHandler);
-			final ByteArrayFileCache resource = o.getSerializedData();
+			final UObject data;
+			try {
+				data = o.getSerializedData() == null ? null : o.getSerializedData().getUObject();
+			} catch (IOException e) {
+				// impossible to test in integration tests, shouldn't occur
+				throw new RuntimeException(
+						"An unexpected error occured: " + e.getLocalizedMessage(), e);
```
ialarmedalien, line 420: s/occured/occurred/

src/us/kbase/workspace/test/kbase/JSONRPCLayerTest.java
MrCreosote: there's not really another way to test this, I think. The current tests are:

* when no external systems are connected to the workspace, the new param does nothing. In that case, there's nothing to query to ensure nothing happened (this is the test above).
* when external systems are connected (in the handle/shock test and the samples test), getting objects is tested with and without the new parameter to ensure it works as expected and the ACLs are updated or not.

I'm not seeing another way to test off the top of my head, anyway

src/us/kbase/workspace/kbase/ArgUtils.java
MrCreosote: fixed

ialarmedalien (commented):
Merge away!

PR #493, opened 2021-11-30
change skip_external_acl_updates to skip_external_system_updates

MrCreosote:
Currently it's all ACL updates, but no saying what those handlers might do in the future.


ialarmedalien (approved):
One of my favourite search-n-replace PRs to date.

PR #495, opened 2022-01-15
Add print stacktrace option to schema updater

MrCreosote:
Rework of https://github.com/kbase/workspace_deluxe/pull/489, which was originally supposed to be a hotfix which turned out to be unnecessary. No point in wasting the code though, so just moving to develop


ialarmedalien (approved):
looks reasonable

PR #497, opened 2022-01-20
Use Sample service bulk ACL update method 

MrCreosote:
Should hopefully speed up setting ACLs for an object with a lot of
sample IDs.

Note that the sample service integration tests required no changes.


python_dependencies/Pipfile
```diff
@@ -13,7 +13,7 @@ requests = "==2.22.0"
 # handle service dependencies
 cachetools = "*"
 configparser = "*"
-pymongo = "*"
+pymongo = "==3.11.0"
```
MrCreosote, line 16: Originally pipenv auto upgraded to v4.X which cause the handle service tests to start failing, so I pinned it and the test magically started working again. 
@Tianhao-Gu heads up

python_dependencies/Pipfile
```diff
@@ -30,6 +30,8 @@ jinja2 = "==2.11.2"
 gevent = "==20.5.0"
 gunicorn = "==20.0.4"
 kafka-python = "==2.0.1"
+greenlet = "==0.4.16"
```
MrCreosote, line 33: This one's a bit weird - if you examine the lock file you'll see that `greenlet` was already installed and all the hashes are the same. However, the new version of the sample service was stuck constantly restarting the workers with a warning about `greenlet` binary compatibility and explicitly installing `greenlet` made it start so *shrug*

dakotablair (approved):
Looks good to me, seems simpler over all.

bio-boris (approved):
LGTM

PR #503, opened 2022-02-16
Trivial code cleanup

MrCreosote:
Did this almost absentmindedly while looking for ways to speed up result tab rendering.

I'm just going straight to develop because I don't care about having an image built from this PR.


ialarmedalien (approved):
LGTM

PR #505, opened 2022-02-17
Unify ObjectIdentifier classes part 1

MrCreosote:
This PR moves the ObjectIdentifier subclasses into the ObjectIdentifier
class proper without any other changes. Future PRs will include:

* Add all subclass methods to superclasses with sane default behavior
* Write a builder to replace the 9 constructors and 2 create methods
that picks the right class to return
* Hide the 2 subclasses and refactor code that uses them
* Write unit tests for the ObjectIdentifier class


ialarmedalien (approved):
🌡️

PR #506, opened 2022-02-17
Add all ObjectIdentifier subclass methods to superclass

MrCreosote:
No need to do bizarre instanceof checking, can now hide the subclass
implementations

Future PRs:
* Write a builder to replace the 9 constructors and 2 create methods
that picks the right class to return
* Hide the 2 subclasses and refactor code that uses them
* Write unit tests for the ObjectIdentifier class


ialarmedalien (approved):
👍

PR #507, opened 2022-02-18
DATAUP-725: Return the object name from the name check method, rename test

MrCreosote:
Better to keep 1 test class per class under test than merge multiple
classes into on test class


ialarmedalien (approved):
More PRs of this length plz!

MrCreosote:
Hmm, you're going to be disappointed by my next PR

PR #509, opened 2022-02-23
DATAUP-725: Add object identifier builder starting from OI

MrCreosote:
Quickly became clear that doing otherwise was going to result in a *lot*
of refactoring that probably wasn't worth it

Handy anyway


ialarmedalien (approved):
Merge away!

PR #510, opened 2022-02-23
DATAUP-725: Add a ObjectIdentifier.Builder copy method.

MrCreosote:
Useful for tests.


ialarmedalien (approved):
looks straightforward enough to lull me into a false sense of security prior to the horrendous PR

PR #513, opened 2022-02-28
DATAUP-725: Switch to primitives in ObjectIdentifier instance vars

MrCreosote:
Now an easy change to make. Assuming a HotSpot JVM with < 32GB of
heap, should reduce 48B to 12B for storing the numbers

Also some minor cleanup.

### DIFF
--- a/src/us/kbase/workspace/database/ObjectIdentifier.java
+++ b/src/us/kbase/workspace/database/ObjectIdentifier.java
@@ -14,7 +14,8 @@ import static us.kbase.common.utils.StringUtils.checkString;
 
 public class ObjectIdentifier {
 	
-	// TODO NOW TEST unittests (move from WorkspaceTest, mostly covered there)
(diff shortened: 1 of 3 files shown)


ialarmedalien (approved):
LGTM

MrCreosote:
The loss of coverage in the `Workspace` and `ObjectIDResolvedWS` classes looks to be due to a change in the order that object IDs from [this line](https://github.com/kbase/workspace_deluxe/blob/17c8ee6677af3b8a0cc1644a34a9949d28c43557/src/us/kbase/workspace/database/Workspace.java#L1890) are run through. Previously it ran through a couple objects before finding the correct object, and now the first object in the loop is the matching object. There's not really any way to ensure this doesn't happen, since the input is a hashset. The order probably changed because `ObjectIdentifier.hashCode()` changed.

So, don't worry about it.

PR #518, opened 2022-03-01
DATAUP-725: Delete unused code in an internal class

MrCreosote:
and make it a package only class, since that's the only place it's used.


ialarmedalien (approved):
👍  to cruft removal!

PR #522, opened 2022-03-02
DATAUP-725: Use Optionals vs nulls in WorkspaceObjectBuilder

MrCreosote:
Avoid the billion dollar mistake

### DIFF
--- a/src/us/kbase/workspace/test/workspace/WorkspaceLongTest.java
+++ b/src/us/kbase/workspace/test/workspace/WorkspaceLongTest.java
@@ -94,10 +94,10 @@ public class WorkspaceLongTest extends WorkspaceTester {
 		
 		//printMem("*** released refs ***");
 		
-		ByteArrayFileCache newdata = ws.getObjects(
+		final ByteArrayFileCache newdata = ws.getObjects(
 				userfoo,
 				Arrays.asList(ObjectIdentifier.getBuilder(bigdataws).withID(1L).build()))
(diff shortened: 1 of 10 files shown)


src/us/kbase/workspace/test/workspace/WorkspaceLongTest.java
```diff
@@ -94,10 +94,10 @@ public void saveWithBigData() throws Exception {
...
 		
-		ByteArrayFileCache newdata = ws.getObjects(
+		final ByteArrayFileCache newdata = ws.getObjects(
 				userfoo,
 				Arrays.asList(ObjectIdentifier.getBuilder(bigdataws).withID(1L).build()))
-				.get(0).getSerializedData();
+				.get(0).getSerializedData().get();
 //		printMem("*** retrieved object ***");
 //		System.gc();
 //		printMem("*** ran gc after retrieve ***");
```
ialarmedalien, line 103: assuming these on your to-delete-at-some-point list

ialarmedalien (approved):
null problems

src/us/kbase/workspace/test/workspace/WorkspaceLongTest.java
MrCreosote: Not really. Part of the point of that test with massive objects is to check for memory usage wrt GC, so those printouts are useful. I suppose I could make a toggle for them rather than have them as comments

PR #524, opened 2022-03-03
DATAUP-725: Add subset data and getter for ObjectInformation to WOD builder

MrCreosote:
* Subset data will be necessary for the rewritten object data pulling code that will take `WorkspaceObjectData.Builder`s as input, since that method will do the subsetting
* Getting the object information from the builder is necessary to get access to the object size in bytes, which will be used to calculate the total size of the requested data and, based on that size, either
    * store it in memory
    * store it on disk
    * throw an error because it's too big

The WOD won't be built until after the data is added, so the size needs to be available from the builder API.


bio-boris (approved):
image

PR #525, opened 2022-03-04
DATAUP-725: Add method to add data to WorkspaceObjectData builder

MrCreosote:
Next step is to return no data from the MWDB.getObjects method and
use this method to populate the builder with data. Then the data size
calculation can be done in the Workspace class and the
ByteArrayFileCacheManager can be initialized up front to save to memory
or disk, getting rid of the thread unsafe dynamic reallocation code.


src/us/kbase/workspace/database/ByteArrayFileCacheManager.java
```diff
@@ -27,6 +27,8 @@
 public class ByteArrayFileCacheManager {
 	
 	//TODO TEST unit tests
+	//TODO JAVADOC
+	// this class is going to get a pretty huge overhaul so wait on tests & javadoc
```
bio-boris, line 31: Ok

bio-boris (approved):
LGTM

PR #526, opened 2022-03-07
DATAUP-725: Use the new data addition method in WS.getObjects()

MrCreosote:
Next: rewrite the BAFCM constructor to just take a boolean that tells it
to store data on disk or in memory, and get rid of the thread unsafe
size calculations.

Seems like the code is quite a bit simpler overall by splitting
MWSDB.getOBjects() into two methods

### DIFF
--- a/src/us/kbase/workspace/database/Workspace.java
+++ b/src/us/kbase/workspace/database/Workspace.java
@@ -60,6 +60,7 @@ import us.kbase.workspace.database.ResourceUsageConfigurationBuilder.ResourceUsa
 import us.kbase.workspace.database.refsearch.ReferenceSearchMaximumSizeExceededException;
 import us.kbase.workspace.database.exceptions.CorruptWorkspaceDBException;
 import us.kbase.workspace.database.exceptions.InaccessibleObjectException;
+import us.kbase.workspace.database.exceptions.NoObjectDataException;
 import us.kbase.workspace.database.exceptions.NoSuchObjectException;
 import us.kbase.workspace.database.exceptions.NoSuchReferenceException;
 import us.kbase.workspace.database.exceptions.NoSuchWorkspaceException;
@@ -120,7 +121,6 @@ public class Workspace {
 		nonNull(listeners, "listeners");
 		noNulls(listeners, "null item in listeners");
 		this.listeners = Collections.unmodifiableList(listeners);
-		db.setResourceUsageConfiguration(rescfg);
 		this.maximumObjectSearchCount = MAX_OBJECT_SEARCH_COUNT_DEFAULT;
 	}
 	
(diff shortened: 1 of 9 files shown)


src/us/kbase/workspace/database/Workspace.java
```diff
@@ -120,7 +121,6 @@ public Workspace(
 		nonNull(listeners, "listeners");
 		noNulls(listeners, "null item in listeners");
 		this.listeners = Collections.unmodifiableList(listeners);
-		db.setResourceUsageConfiguration(rescfg);
```
MrCreosote, line 123: The DB layer no longer needs access to the resource config since the data size calcs happen in the business layer. There are several changes around this

src/us/kbase/workspace/database/Workspace.java
```diff
@@ -1068,129 +1067,96 @@ public UserWorkspaceIDs listWorkspaceIDs(
...
-			stddata.clear();
-			removeInaccessibleDataCopyReferences(user, toProc);
-			// TODO CODE return optionals instead of nulls
-			return toProc.stream().map(b -> b == null ? null : b.build())
-					.collect(Collectors.toList());
-		} catch (RuntimeException | Error | CorruptWorkspaceDBException |
-				WorkspaceCommunicationException | NoSuchObjectException |
-				TypedObjectExtractionException e) {
-			destroyGetObjectsResources(stddata);
-			destroyGetObjectsResources(refdata);
```
MrCreosote, line 1127: Since fetching the data is the last thing that happens in this method and the DB layer data fetching method cleans up after itself if things go wrong, there's no longer any need for this cleanup code

src/us/kbase/workspace/test/workspace/WorkspaceTest.java
```diff
@@ -8699,10 +8699,10 @@ public void maxReturnedObjectSize() throws Exception {
 			List<ObjectIdentifier> all = new LinkedList<ObjectIdentifier>();
 			all.addAll(ois1l2);
 			all.addAll(bothoi);
-			ws.setResourceConfig(build.withMaxReturnedDataSize(60).build());
+			ws.setResourceConfig(build.withMaxReturnedDataSize(80).build());
 			destroyGetObjectsResources(ws.getObjects(user, all));
-			ws.setResourceConfig(build.withMaxReturnedDataSize(59).build());
-			err = new IllegalArgumentException(String.format(errstr, 60, 59));
+			ws.setResourceConfig(build.withMaxReturnedDataSize(79).build());
+			err = new IllegalArgumentException(String.format(errstr, 80, 79));
```
MrCreosote, line 8705: This change actually fixed a bug and so this test had to be adjusted. Previously, if a user requested data from the same object twice (which is possible - for example `myws/myobj` and `1/1/1` could point to the same object) it wasn't counted against the amount of data returned because the data was only pulled from the `BlobStore` once. Of course, the data is duplicated in the return and so the actual data package could've been larger than the 1GB limit. Now the data size is calculated based on the data returned to the user.

bio-boris (approved):
LGTM

PR #527, opened 2022-03-10
DATAUP-725: Rewrite ByteArrayFileCacheManager for thread safety

MrCreosote:
In the current implementation, BAFCM keeps a running tally of the size
of data in memory and when the limit on that size is exceeded it dumps the current
file to disk. It updates the tally after each file is completed, which
assumes only one file is being processed at a time.

This is not safe for the case where data creation is parallelized and
multiple data objects are streaming into memory simultaneously. This
commit rewrites BAFCM to simply put all data into memory or disk,
depending on the wishes of the creator of the BAFCM. The Workspace class
is the creator, which chooses to use disk storage if the size of the
incoming data + potential subsets is greater than the configured maximum
allowed.

This does have a drawback that if you have large data but small subsets,
the data may be stored on disk when it might have fit in memory in the
prior implementation. However, this is a very specific case for a very
specific amount of data that is close to the in memory limit. If there
are no subsets, the subsets are close to the size of the objects, the
data exceeds the in memory limit, or the data is not close to the in
memory limit there's no effective difference.

An alternative implementation could keep track of all simultaneous
incoming data byte by byte, but that seems like a great deal of
complication and source of parallelization bugs to fix a relatively
small proportion of cases.

TODO: Javadoc for BAFCM


src/us/kbase/workspace/database/ByteArrayFileCacheManager.java
```diff
@@ -1,5 +1,7 @@
 package us.kbase.workspace.database;
```
MrCreosote, line 1: This rewrite is so sweeping it might be easier to just look at the new file rather than the diff, or just get a sense of how the old implementation worked from the diff or prior commit.

src/us/kbase/workspace/database/ByteArrayFileCacheManager.java
```diff
@@ -269,18 +167,15 @@ public long getSize() {
...
+		// in docs note that this method or parent.getUObject() must complete before calling either
+		// one again or exceptions will be thrown. Also note not thread safe.
+		// fixing this bug would be a lot of work and not necessary at the moment
 		public UObject getUObject() throws JsonParseException, IOException {
 			checkIfDestroyed();
 			jts.setRoot(null);
 			return new UObject(jts);
 		}
 		
-		public JsonNode getAsJsonNode()
```
MrCreosote, line 278: Completely unused

src/us/kbase/workspace/database/ByteArrayFileCacheManager.java
```diff
@@ -338,7 +237,7 @@ public void destroy() {
 			} catch (IOException ioe) {
 				//nothing can be done
 			}
-			if (tempFile != null && tempFile.exists()) {
+			if (tempFile != null) {
 				tempFile.delete();
```
MrCreosote, line 241: If the tempfile doesn't exist `delete()` is a noop, so no need to check for existence

bio-boris (approved):
image

bio-boris:
>An alternative implementation could keep track of all simultaneous
incoming data byte by byte, but that seems like a great deal of
complication and source of parallelization bugs to fix a relatively
small proportion of cases.

So would you have to have a counter that multiple threads would be incrementing?

MrCreosote:
Right

PR #538, opened 2022-03-11
DATAUP-725: Parallelize fetching data from the blob backend (#8/11)

MrCreosote:
TODO:
* Hook up the dynamic config to the parallelization code
* Add documentation

### DIFF
--- a/src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
+++ b/src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
@@ -33,6 +33,12 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 import org.bson.Document;
 import org.bson.types.ObjectId;
@@ -135,6 +141,8 @@ public class MongoWorkspaceDB implements WorkspaceDatabase {
 	// could cause index build fail and corrupt data in the db
 	// Need to look into this before upgrading to 4.2
 
+	private static final long OBJECT_DATA_FETCH_TIMEOUT_SEC = 900;
+	
 	public static final AllUsers ALL_USERS = Workspace.ALL_USERS;
 
 	//TODO CONFIG this should really be configurable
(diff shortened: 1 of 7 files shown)


src/us/kbase/workspace/test/database/mongo/MongoWorkspaceDBTest.java
```diff
@@ -623,6 +661,93 @@ public Exception failAddDataToObjects(
...
+		 * a higher scaling factor should result in a faster return, which is not something
+		 * that should be in automated tests as it's heavily dependent on the environment.
+		 * Furthermore, in automated tests like this where network speed is not an issue
+		 * little to no benefit will be seen from parallelizing the requests as IO should be
+		 * local disk bound.
+		 * 
+		 * Hence, this test just tries different scaling factors and prints out the speed
+		 * for informational purposes. Any rigorous speed testing should take place outside the
+		 * context of unit or even integration tests.
+		 */
```
MrCreosote, line 678: All that said, even in this unrealistic test, where I'd expect the impact of threading to be the least, it has a pretty significant impact. This is the result of the print statement at the end of this test:
```
{1=4619, 2=2207, 5=1103, 10=848, 50=832, 100=876, 500=1232}
```

src/us/kbase/workspace/database/mongo/MongoWorkspaceDB.java
```diff
@@ -2308,80 +2316,139 @@ private void saveData(
...
-							info.getWorkspaceId(), info.getObjectId(), info.getVersion()), e);
-				}
-				for (final WorkspaceObjectData.Builder b: chksum2obj.get(chksum)) {
-					/* might be subsetting the same object the same way multiple times, but
-					 * probably unlikely. If it becomes a problem memoize the subset
-					 */
-					b.withData(getDataSubSet(data, b.getSubsetSelection(), dataManager));
-				}
+			final List<Future<Void>> futures = eserv.invokeAll(
+					callables, OBJECT_DATA_FETCH_TIMEOUT_SEC, TimeUnit.SECONDS);
```
MrCreosote, line 2345: An alternative implementation could be to have a global executor service and submit `backendScaling` meta jobs that all share a synchronized list of jobs and pull jobs off the queue in a loop. That would improve over the current implementation in two ways:

* The single executor service would allow capping the total number of threads talking to the backend per service and could be configured (probably in `deploy.cfg`, switching executors live seems iffy). Currently the effective number of threads is the server thread count (which is 500 for tomcat IIRC) * `backendScaling`.
* The meta jobs could cancel early if one job fails, and could even set a semaphore so all the meta jobs could fail. Currently all jobs run to completions regardless of the status of the other jobs.

All that being said, the current implementation is much simpler and probably perfectly fine, so YAGNI.

PR #549, opened 2022-03-18
Delete or organize non-core files

MrCreosote:
This has needed to be done for a long time...

Removed no longer useful files, consolidated other files that aren't
core code under `performanceanddevelopment`. Hopefully this will make
the codebase a bit easier to understand


hooks/build
```diff
@@ -1,16 +0,0 @@
-#!/bin/bash
```
MrCreosote, line 1: dockerhub build cruft

Makefile
```diff
@@ -69,10 +69,7 @@ compile-typespec:
 		$(SERVICE).spec
 	rm lib/biokbase/workspace/authclient.py
 
-test: test-client test-service
-
-test-client: test-service
-	$(ANT) test_client_import
```
MrCreosote, line 75: Testing that the client built correctly made sense back when the Workspace was the first and only Java SDK service. Now it's a waste of time

src/AllWorkspaceTestsRunner.java
```diff
@@ -1,76 +0,0 @@
-import java.io.File;
```
MrCreosote, line 1: Unused, you can do the same thing with patterns in ant anyway (and presumably Gradle, etc.)

bio-boris (approved):

MrCreosote:
LGTM and the docker build were already successful for this PR, changing the base to develop caused them to be rerun

PR #550, opened 2022-03-19
Add a Blobstore controller

MrCreosote:
...to replace the Shock controller


MrCreosote:
LGTM and the docker build were already successful for this PR, changing the base to develop caused them to be rerun

PR #560, opened 2022-03-29
PTV-1773: Clean up external data unit section of provenance

MrCreosote:
Make class immutable, add a builder.

Also a lot of cleanup in other areas of the code related to provenance,
and marked some TODOs for future PRs around better error handling and
testing thereof.

Since EDU will now reject malformed URLs, I pulled all the URLs from the
production WS DB and ran them through `.withDataURL()` and all of them
were fine.


lib/biokbase/workspace/client.py
```diff
@@ -791,7 +791,8 @@ def save_objects(self, params, context=None):
            data unit. A piece of data from a source outside the Workspace. On
```
MrCreosote, line 791: generated

lib/Bio/KBase/workspace/Client.pm
```diff
@@ -9423,6 +9423,8 @@ Workspace.
 On input, only one of the resource_release_date or
```
MrCreosote, line 9423: generated

src/us/kbase/workspace/ExternalDataUnit.java
```diff
@@ -18,6 +18,7 @@
  * Workspace.
```
MrCreosote, line 18: generated

src/us/kbase/workspace/kbase/ArgUtils.java
```diff
@@ -113,6 +113,7 @@ public static Provenance processProvenance(final WorkspaceUser user,
 			return p;
 		}
 		for (final ProvenanceAction a: actions) {
+			// TODO PROV what if a is null?
```
MrCreosote, line 116: There's a number of these `TODO PROV`s around. Normally I'd take care of them in the same PR, but this PR is pretty huge already, so I'm putting them off until later PRs.

workspace.spec
```diff
@@ -427,6 +427,8 @@ module Workspace {
 		On input, only one of the resource_release_date or
 		resource_release_epoch may be supplied. Both are supplied on output.
 		
+		All fields are optional, but at least one field much be present.
```
ialarmedalien, line 430: `s/much/must/`

MrCreosote: *&*)&&*^) fixed

src/us/kbase/workspace/test/kbase/JSONRPCLayerTester.java
```diff
@@ -573,16 +573,26 @@ public boolean equals(Object obj) {
...
-			.withNoData(1L)
-			.withObjects(Arrays.asList(toObjSpec(id)))).getData().get(0);
+		final ObjectData ret1p;
+		try {
+			ret1p = CLIENT1.getObjects2(new GetObjects2Params()
+					.withNoData(1L)
+					.withObjects(Arrays.asList(toObjSpec(id)))).getData().get(0);
+		} catch (ServerException e) {
+			System.err.println(e.getData());
+			throw e;
```
ialarmedalien, line 594: Will this throw and abort the rest of the tests in this function? Is there a reason why this one has the try/catch? If this `getObjects2` fails, will all the following tests/invocations of `getObjects2` also fail?

MrCreosote: > Will this throw and abort the rest of the tests in this function?

yes

> Is there a reason why this one has the try/catch?

The `toString` method of `ServerException` doesn't include the `data` field (and def shouldn't), which is the server side stack trace and so makes debugging the issue much easier. This particular call was failing server side so I added the try catch to get the server side stack trace. Since it doesn't hurt anything to leave it there, I did that.

> If this getObjects2 fails, will all the following tests/invocations of getObjects2 also fail?

Isn't this the same as the first question? If not can you elaborate as I'm a bit fogged re the difference

ialarmedalien: It's the same Q phrased very slightly differently to befuddle and confuse.

ialarmedalien (approved):
👍

PR #570, opened 2022-04-07
Print a stacktrace when the server can't start

MrCreosote:
Slightly hacky but fine in the startup code


MrCreosote:
Really need to take another look at logging anyway, I assume it should be switched from syslog to sysout/err

jsfillman (approved):
Let's :shipit: !

PR #584, opened 2022-04-19
PTV-1786: Separate the admin commands from the admin framework

MrCreosote:
Makes it easier to run with different sets of admin commands or different implementations of the same set

Step two of this process is making a WorkspaceAdministrator builder and using that instead of just taking a map of admin commands in the constructor. At that point I'll also add new WA unit tests.

### DIFF
--- a/src/us/kbase/workspace/kbase/admin/AdministrationCommandSetBuilder.java
+++ b/src/us/kbase/workspace/kbase/admin/AdministrationCommandSetBuilder.java
@@ -1,0 +1,486 @@
+package us.kbase.workspace.kbase.admin;
+
+import static us.kbase.workspace.kbase.ArgUtils.wsInfoToTuple;
+import static us.kbase.workspace.kbase.IdentifierUtils.processWorkspaceIdentifier;
+
+import java.io.IOException;
+import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


src/us/kbase/workspace/kbase/admin/AdministrationCommandSetBuilder.java
```diff
@@ -0,0 +1,486 @@
+package us.kbase.workspace.kbase.admin;
```
MrCreosote, line 1: The essence of the PR is moving the individual admin commands from WorkspaceAdministration to this builder so eventually the builder can install alternate sets of commands depending the needs of the service.

src/us/kbase/workspace/kbase/admin/WorkspaceAdministration.java
```diff
@@ -200,304 +268,12 @@ public Object runCommand(
 			}
 			throw ioe;
 		}
-		// should look into some sort of interface w/ registration instead of a massive if else
```
MrCreosote, line 203: I think this has been thoroughly looked into

ialarmedalien (approved):
thumbs up

PR #587, opened 2022-04-21
PTV-1786: Rename the command set installer

MrCreosote:
More accurately describes what it does


ialarmedalien (approved):
I guess this is OK...

MrCreosote:
Your lack of enthusiasm is really hurtful to me

PR #588, opened 2022-04-22
PTV-1786: Extract type methods from the server class into their own class

MrCreosote:
Will allow creating an interface and alternate implementation that delegates to another workspace service

### DIFF
--- a/src/us/kbase/workspace/kbase/LocalTypeServerMethods.java
+++ b/src/us/kbase/workspace/kbase/LocalTypeServerMethods.java
@@ -1,0 +1,323 @@
+package us.kbase.workspace.kbase;
+
+import static us.kbase.common.utils.ServiceUtils.checkAddlArgs;
+import static us.kbase.workspace.kbase.ArgUtils.longToBoolean;
+
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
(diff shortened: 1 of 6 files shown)


src/us/kbase/workspace/kbase/LocalTypeServerMethods.java
```diff
@@ -0,0 +1,323 @@
+package us.kbase.workspace.kbase;
```
MrCreosote, line 1: This PR is effectively just moving all the code from `WorkspaceServer` that deals with the type database in here, almost verbatim, plus the two methods from `WorkspaceServerMethods`.

src/us/kbase/workspace/kbase/LocalTypeServerMethods.java
```diff
@@ -0,0 +1,323 @@
...
+		}
+		final List<String> add = params.getNewTypes() != null ?
+				params.getNewTypes() : new ArrayList<String>();
+		final List<String> rem = params.getRemoveTypes() != null ?
+				params.getRemoveTypes() : new ArrayList<String>();
+		final Map<String, Long> deps = params.getDependencies() != null ?
+				params.getDependencies() : new HashMap<String, Long>();
+		final Map<TypeDefName, TypeChange> res;
+		if (params.getMod() != null) {
+			 res = types.compileTypeSpec(
```
ialarmedalien, line 112: is this going to be covered somewhere else / in a future PR?

MrCreosote: Not in the near future anyway. It hasn't been covered for the last 8 years apparently, so for now I'm leaving things as is.

ialarmedalien (approved):
Looks fine

PR #590, opened 2022-04-22
PTV-1786: Add a small workspace delegation helper

MrCreosote:
Hides some of the annoyances of the workspace client API.

Also some minor cleanup


src/us/kbase/workspace/kbase/LocalTypeServerMethods.java
```diff
@@ -114,12 +115,12 @@ public Map<String,String> registerTypespec(
 		final Map<TypeDefName, TypeChange> res;
 		if (params.getMod() != null) {
 			 res = types.compileTypeSpec(
-					getUser(authPart), params.getMod(),
+					getUser(token), params.getMod(),
```
ialarmedalien, line 118: will you add a test for this?

ialarmedalien (approved):
Looks fine

src/us/kbase/workspace/kbase/LocalTypeServerMethods.java
MrCreosote: Not right now. This is the same code we talked about before, where I just moved it from the server class into this class verbatim so I could write an alternate implementation.

ialarmedalien: Got it, cool.

MrCreosote:
The difference between 42e6d52 and d9bcf5c is just merge commits so the code coverage is accurate. Not sure why codecov didn't pick up the new builds.

PR #597, opened 2022-04-28
PTV-1786: Implement TypeServerMethods that delegate to another workspace

MrCreosote:
For workspace configs that provide a URL for another workspace, this'll be used instead of LocalTypeServerMethods.


src/us/kbase/workspace/kbase/DelegatingTypeServerMethods.java
```diff
@@ -0,0 +1,158 @@
...
+			final AuthToken token,
+			final boolean asAdmin) // admin methods should not call this when delegating
+			throws WorkspaceDelegationException {
+		delegator.delegate(token, c -> {c.removeModuleOwnership(params); return null;});
+	}
+
+	@Override
+	public void requestModuleOwnership(final String mod, final AuthToken token)
+			throws WorkspaceDelegationException {
+		delegator.delegate(token, c -> {c.requestModuleOwnership(mod); return null;});
```
briehl, line 59: Why do these specifically return `null`? Shouldn't they naturally return `null` if structured as below, since the delegated method returns `null`? Or would they return some kind of empty container?

briehl (changes requested):
Mainly just a little question. Everything else seems pretty straightforward and looks fine. 👍

src/us/kbase/workspace/kbase/DelegatingTypeServerMethods.java
MrCreosote: I had a lot of fun figuring this out. I'm not totally sure I 100% grok it but this is my understanding of what's going on.

The `delegate` method takes a `WorkspaceCommand<T>`, which is a functional interface with a parameterized return type. That means any lambda functions passed as a `WorkspaceCommand` *must* return something. The 3 methods here that are suffixed with an explicit return statement all return `void`, which means they return nothing - not even `null`. If you take out those return statements you get a compiler error complaining that it can't figure out the type of the return value of the lambda function.

TLDR - if you take out the explicit return the code doesn't compile.

briehl: I get it. That's annoying, but I get it.

briehl (approved):
👍

PR #600, opened 2022-05-05
PTV-1786: Simplify type provider interface

MrCreosote:
The absolute type def and the jsonschema can be fetched in one call to
the workspace, so make the provider a single call.

### DIFF
--- a/src/us/kbase/typedobj/core/LocalTypeProvider.java
+++ b/src/us/kbase/typedobj/core/LocalTypeProvider.java
@@ -1,5 +1,7 @@
 package us.kbase.typedobj.core;
 
+import static java.util.Objects.requireNonNull;
+
 import us.kbase.typedobj.db.TypeDefinitionDB;
 import us.kbase.typedobj.exceptions.NoSuchModuleException;
 import us.kbase.typedobj.exceptions.NoSuchTypeException;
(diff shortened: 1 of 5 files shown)


src/us/kbase/typedobj/core/LocalTypeProvider.java
```diff
@@ -1,42 +1,32 @@
...
+import static java.util.Objects.requireNonNull;
+
 import us.kbase.typedobj.db.TypeDefinitionDB;
 import us.kbase.typedobj.exceptions.NoSuchModuleException;
 import us.kbase.typedobj.exceptions.NoSuchTypeException;
 import us.kbase.typedobj.exceptions.TypeStorageException;
 
 /** A type provider for the typed object validator that takes a direct instance
  * of a type database.
- * @author gaprice@lbl.gov
```
ialarmedalien, line 10: they'll still be able to find you through the git blame

ialarmedalien (approved):

src/us/kbase/typedobj/core/LocalTypeProvider.java
MrCreosote: It's more about futilely trying to avoid email harvesting bots

PR #605, opened 2022-05-18
PTV-1786: Add type fetch exception to type provider

MrCreosote:
In preparation for delgating type provider implementation

### DIFF
--- a/src/us/kbase/typedobj/core/TypeProvider.java
+++ b/src/us/kbase/typedobj/core/TypeProvider.java
@@ -9,6 +9,26 @@ import us.kbase.typedobj.exceptions.TypeStorageException;
 /** Provides type definitions to the TypedObjectValidator. */
 public interface TypeProvider {
 
+	/** An error thrown when type information could not be fetched from a remote resource. */ 
+	public static class TypeFetchException extends Exception {
+		private static final long serialVersionUID = 1L;
+
+		/** Create the exception.
+		 * @param message the exception message.
+		 */
+		public TypeFetchException(final String message) {
+			super(message);
+		}
+
+		/** Create the exception.
+		 * @param message the exception message.
+		 * @param cause the cause of the exception.
+		 */
+		public TypeFetchException(final String message, final Throwable cause) {
+			super(message, cause);
+		}
+	}
+	
 	/** A resolved type and its JSONSchema. */
 	public static class ResolvedType {
 		private final AbsoluteTypeDefId type;
@@ -70,15 +90,17 @@ public interface TypeProvider {
 	}
 	
 	/** Retrieves the resolved type and JsonSchema type definition document for the
-	 * specified type.
+	 * specified type. Unreleased types are not returned unless the type definition ID is
+	 * absolute.
 	 * @param type a type id.
 	 * @return the resolved type with JSONSchema.
 	 * @throws TypeStorageException if an error occurs with the type storage
 	 * engine.
 	 * @throws NoSuchModuleException if the module for the type does not exist.
 	 * @throws NoSuchTypeException if the type does not exist.
+	 * @throws TypeFetchException if an error occurred while fetching the type.
 	 */
 	public ResolvedType getTypeJsonSchema(final TypeDefId type)
 			throws NoSuchTypeException, NoSuchModuleException,
-			TypeStorageException;
+				TypeStorageException, TypeFetchException;
 }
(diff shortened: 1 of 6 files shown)


src/us/kbase/workspace/database/Workspace.java
```diff
@@ -918,32 +920,33 @@ private ValidatedTypedObject validate(
...
-			TypedObjectValidationException {
+				TypedObjectValidationException {
 		final ValidatedTypedObject rep;
 		try {
 			rep = validator.validate(wo.getData(), wo.getType(), idhandler);
 		} catch (NoSuchTypeException nste) {
 			throw new TypedObjectValidationException(String.format(
-					"Object %s failed type checking:\n",
-					getObjectErrorId(wo, objcount))
+					"Object %s failed type checking:\n", getObjectErrorId(wo, objcount))
```
MrCreosote, line 929: I get rid of these newlines in a future PR

src/us/kbase/workspace/database/Workspace.java
```diff
@@ -918,32 +920,33 @@ private ValidatedTypedObject validate(
...
+					"- a fatal JSON processing error occurred: " + jpe.getMessage(), jpe);
 		} catch (IOException ioe) {
 			throw new TypedObjectValidationException(String.format(
-					"A fatal IO error occured while type checking object %s: ",
-					getObjectErrorId(wo, objcount)) + ioe.getMessage(), ioe);
+					"Object %s failed type checking ", getObjectErrorId(wo, objcount)) + 
+					"- a fatal IO error occurred: " + ioe.getMessage(), ioe);
+		} catch (TypeFetchException tfe) {
+			throw new TypedObjectValidationException(String.format(
+					"Object %s failed type checking ", getObjectErrorId(wo, objcount)) +
```
MrCreosote, line 947: Hmm, now I'm thinking I should change `getObjectErrorId` to `getExceptionPrefix` and include the first string since it's always the same for this block

src/us/kbase/workspace/test/workspace/WorkspaceUnitTest.java
```diff
@@ -516,4 +531,169 @@ private void failGetObjects(
...
+				id,
+				SAVE_OBJECT_FAIL_BAD_DATA,
+				TypeDefId.fromTypeString(SAVE_OBJECT_FAIL_VAL_TYPE_STRING),
+				new WorkspaceUserMetadata(),
+				Provenance.getBuilder(new WorkspaceUser("u"), now()).build(),
+				false);
+	}
+	@Test
+
+	public void saveObjectValidationFailNoSuchModuleException() throws Exception {
```
MrCreosote, line 558: This messed up spacing is fixed locally and will be in a future PR

src/us/kbase/typedobj/core/TypeProvider.java
```diff
@@ -9,6 +9,26 @@
...
+			super(message);
+		}
+
+		/** Create the exception.
+		 * @param message the exception message.
+		 * @param cause the cause of the exception.
+		 */
+		public TypeFetchException(final String message, final Throwable cause) {
+			super(message, cause);
+		}
```
ialarmedalien, line 29: is this particularly tricky to write a test for? Or are tests coming in a future PR?

src/us/kbase/workspace/test/workspace/WorkspaceUnitTest.java
ialarmedalien: can you set up a code linter/formatter to run as a pre-commit hook so you don't have to worry about this kind of thing?

src/us/kbase/workspace/database/Workspace.java
ialarmedalien: could always put that in the next PR

MrCreosote: done, although it's more like 10 PRs down the line

src/us/kbase/workspace/test/workspace/WorkspaceUnitTest.java
MrCreosote: Hmm, I'm not really a huge fan of style linters - 99% of the stuff they catch is trivial and not worth worrying about. There's just too much noise, and the number of times linters have actually helped me is tiny.

These cases are rare enough and harmless enough that I don't think the trouble is worth it. Maybe my perspective will change in the future, but for right now I think the cure is worse than the disease.

src/us/kbase/typedobj/core/TypeProvider.java
MrCreosote: Future PR. I used to explicitly test exceptions but the code is so trivial it seems pointless, and it's better to test it by using it in a test for a larger piece of code.

ialarmedalien (approved):

PR #613, opened 2022-05-23
PTV-1786: Simplify type operation delegation part 1/2

MrCreosote:
Effectively combines DelegatingTypeServerMethods and WorkspaceDelegator into one class. This has 2 benefits:

1) It makes it much simpler to switch to a TypeServiceClient when that exists since you only have to worry about changing one class
2) It makes the tests much simpler since all the delegation functional interface stuff is hidden

Next step is replacing the 2 classes with the new class.

### DIFF
--- a/src/us/kbase/workspace/kbase/TypeClient.java
+++ b/src/us/kbase/workspace/kbase/TypeClient.java
@@ -1,0 +1,254 @@
+package us.kbase.workspace.kbase;
+
+import static java.util.Objects.requireNonNull;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.List;
+import java.util.Map;
+
+import us.kbase.auth.AuthToken;
+import us.kbase.common.service.JsonClientException;
+import us.kbase.common.service.ServerException;
+import us.kbase.common.service.UObject;
+import us.kbase.common.service.UnauthorizedException;
+import us.kbase.workspace.FuncInfo;
+import us.kbase.workspace.GetModuleInfoParams;
+import us.kbase.workspace.GrantModuleOwnershipParams;
+import us.kbase.workspace.ListAllTypesParams;
+import us.kbase.workspace.ListModuleVersionsParams;
+import us.kbase.workspace.ListModulesParams;
+import us.kbase.workspace.ModuleInfo;
+import us.kbase.workspace.ModuleVersions;
+import us.kbase.workspace.RegisterTypespecCopyParams;
+import us.kbase.workspace.RegisterTypespecParams;
+import us.kbase.workspace.RemoveModuleOwnershipParams;
+import us.kbase.workspace.TypeInfo;
+import us.kbase.workspace.WorkspaceClient;
+
+/** An type client that delegates type operations to another service.
+ * Note that any asAdmin param toggles are ignored - administration methods should be delegated
+ * to {@link #administer(UObject, AuthToken)} rather than the standard methods.
+ */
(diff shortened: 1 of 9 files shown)


src/us/kbase/workspace/kbase/TypeClient.java
```diff
@@ -0,0 +1,254 @@
...
+import us.kbase.workspace.ListModulesParams;
+import us.kbase.workspace.ModuleInfo;
+import us.kbase.workspace.ModuleVersions;
+import us.kbase.workspace.RegisterTypespecCopyParams;
+import us.kbase.workspace.RegisterTypespecParams;
+import us.kbase.workspace.RemoveModuleOwnershipParams;
+import us.kbase.workspace.TypeInfo;
+import us.kbase.workspace.WorkspaceClient;
+
+/** An type client that delegates type operations to another service.
```
MrCreosote, line 29: s/An/A/ locally

src/us/kbase/workspace/test/kbase/TypeClientTest.java
```diff
@@ -0,0 +1,462 @@
...
+		}
+	}
+	
+	@Test
+	public void grantModuleOwnership() throws Exception {
+		for (final boolean asAdmin: list(true, false)) {
+			// admin should make no difference
+			final TestMocks m = initTestMocks(new URL("https://whee1111.com"));
+			
+			when(m.wcp.getClient(new URL("https://whee1111.com"))).thenReturn(m.wc);
```
ialarmedalien, line 111: YMMV, but I would find it easier and quicker to parse these tests if you set up two variables representing an `https://` URL and an `http://` URL respectively and then used those throughout.

ialarmedalien: I do appreciate that much of the fun in writing the tests would be eliminated by such an URL scheme, though.

ialarmedalien (approved):
Up to you whether you want to take action on the URL comment or not.

src/us/kbase/workspace/test/kbase/TypeClientTest.java
MrCreosote: I tend to prefer to change up variable values among similar tests when it's simple to do so just in case there's some sort of weird behavior that only shows up for a subset of input values. 

It also catches the old
```
def my_func(arg1, arg2):
    return 42 // make tests pass
```

src/us/kbase/workspace/test/kbase/TypeClientTest.java
```diff
@@ -0,0 +1,462 @@
...
+		
+	}
+
+	@Test
+	public void getFuncInfo() throws Exception {
+		final TestMocks m = initTestMocks(new URL("https://gfi.com"));
+		
+		final FuncInfo fi = new FuncInfo().withDescription("some function or other");
+		when(m.wcp.getClient(new URL("https://gfi.com"))).thenReturn(m.wc);
+		when(m.wc.getFuncInfo("funcky chicken")).thenReturn(fi);
```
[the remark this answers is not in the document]
MrCreosote: function + funky = funcky

src/us/kbase/workspace/test/kbase/TypeClientTest.java
MrCreosote: And honestly IMO you need to inject as much fun as possible into writing tests to even slightly offset the soul crushing drudgery of writing them

PR #618, opened 2022-05-25
Make the server log to stdout

MrCreosote:
Also ignore the KB_SERVICE_NAME variable


MrCreosote:
On reflection, this PR probably needs some extra explanation. Up until now, `JsonServerServlet`, the superclass of `WorkspaceServer`, set up the authentication and logging for the system before any of the `WorkspaceServer` initialization occurred. Unfortunately, it also used those loggers before `WorkspaceServer` could touch them and they're set up to log to syslog. Hence, all of that setup needs to be done by the `WorkspaceServer` class instead to ensure that all logs go to stdout, which is why there's all the new code for getting the config, setting up the loggers, and setting up the auth client.

As usual for server startup code, the unhappy path code is a real pain to test and for now is tested manually. Long term I'd like to embed the `WorkspaceServer` class in a more general HTTP server framework and just call a constructor to set it up rather than having the Servlet API set it up, which might make things a bit easier to test, although given how infrequently this code changes it still might not be worth it.

However, the new set up does allow getting rid of the old auth server url config param, which is nice. It also means that there's only one auth client instance now, instead of one for `JsonServerServlet` and one for everything else.

src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
```diff
@@ -795,6 +796,8 @@ public boolean equals(Object obj) {
 				return false;
 		} else if (!delegateTypeTarget.equals(other.delegateTypeTarget))
 			return false;
+		if (dontTrustXIPHeaders != other.dontTrustXIPHeaders)
```
ialarmedalien, line 799: presumably `dontTrustXIPHeaders` can never be `null`, hence no null test here

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
```diff
@@ -521,8 +517,8 @@ public void configMaximalS3AndDelegateTypes() throws Exception {
 				"type-delegation-target=http://localhost:12345\n" +
 				"mongodb-retrywrites=true\n" +
 				"mongodb-user=muser\n" +
-				"auth-service-url=" + AUTH_LEGACY_URL + "\n" +
 				"auth2-service-url=" + CI_SERV + "auth\n" +
+				"dont-trust-x-ip-headers=true\n" +
```
ialarmedalien, line 521: do you need a corresponding test with `"dont_trust_x_ip_headers=true\n"`? I couldn't see one below.

ialarmedalien (approved):
Approving on the assumption that if there is anything missing, you'll add it to the next PR.

src/us/kbase/workspace/kbase/KBaseWorkspaceConfig.java
MrCreosote: Right, it's a `boolean` so always either `true` or `false`

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
MrCreosote: I'm guessing you missed these lines maybe? https://github.com/kbase/workspace_deluxe/pull/618/files#diff-4b9e468fe59a05ef0feeaaa0f80dcfbbaf63767992679b8ae04a1be581cd9244R575-R609

MrCreosote: If you're talking about not having it printed in the param report, the legacy key is really old and has never been documented and I suspect is not used at all, so I didn't bother to put it in the report. I don't think this feature is used at all, period, actually, but it was less work to support it that to confirm that.

MrCreosote:
I don't have any more PRs in fact (yay!) so I'll ping you for a rereview

src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java
ialarmedalien: I was talking about the printed version -- your explanation makes sense so 👍

PR #627, opened 2022-06-17
Add new gha and remove old gha

bio-boris:
This removes old GHA 
Adds new GHA
Builds now are created upon MERGE to Develop (workspace-deluxe-develop:latest)
Builds now are created upon pressing the "RELEASE" button and tagging a release and adding a description (workspace-deluxe:0.0.X)


MrCreosote:
Looks like Trivy is picking up a ton of stuff from the base image. What's the expectation here?

MrCreosote:
It also looks like the `docker_build` task is blocked or something

.github/workflows/pr_build.yml
```diff
@@ -0,0 +1,43 @@
...
+  pull_request:
+    branches:
+      - develop
+      - main
+      - master
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
```
MrCreosote, line 13: Why do builds on a close?

.github/workflows/manual-build.yml
```diff
@@ -0,0 +1,11 @@
+---
```
MrCreosote, line 1: What does this action do? Could you add some docs?

MrCreosote (changes requested):
see comments above

.github/workflows/pr_build.yml
bio-boris: Not sure, maybe it should be "merged" instead?

.github/workflows/manual-build.yml
bio-boris: It allows you to build an image on demand by manually running a workflow from the github actions tab. The docs will be added at some point to https://github.com/kbase/.github/tree/main/guide

bio-boris:
The expectation with trivy is to alert for CVEs and then give the option for the developer to address these issues to lower  the amount of alerts. I believe it should also be looking for `CVE-2021-4104 CVE-2021-44228 CVE-2021-45046 CVE-2022-22965` and if those are present, then we cannot deploy these to any resources hosted at DOE Infrastructure. @jsfillman May have some additional comments.

MrCreosote:
Ok, so for Trivy, it sounds like there's no particular expectations for the dev regarding timeline for closing, etc. It's just informational, and if we can close issues that's great but not necessarily expected.

For most of these I'd have to make a whole new image that doesn't depend on the severely outdated `kb_jre`

.github/workflows/pr_build.yml
bio-boris: I'm just updating active repos to the lastest version of the DevOps scripts by pressing the button to add workflows. It works on closed, but the goal is to get rid of these old scripts and use the new scripts on repos that use the new actions.

bio-boris: I think I got confused, “merged” is not a valid option.. So yea close is the only way to trigger a build on merge I think I got confused, “merged” is not a valid option

MrCreosote:
I don't think so, I think the only open issue at this point is confirming my impression of the trivy stuff is correct

bio-boris:
> Ok, so for Trivy, it sounds like there's no particular expectations for the dev regarding timeline for closing, etc. It's just informational, and if we can close issues that's great but not necessarily expected, right?
> 
> For most of these I'd have to make a whole new image that doesn't depend on the severely outdated `kb_jre`

Yes no action required for now but will possibly be required in the future

MrCreosote (approved):

PR #630, opened 2022-06-21
Bump version to 0.14.1

MrCreosote:


MrCreosote:
I usually don't put stuff in release notes that doesn't affect the user or server administrators. I normally wouldn't put in the S3 client changes due to that but there's nothing else in this release.

If you really feel strongly about it, I don't have super strong feelings and can do another PR

PR #635, opened 2022-10-19
CE-93: add organization class

ialarmedalien:
Adding in the Organisation object that will be used in the credit metadata.

Per the schema:

organizationName: string, required
organizationID: string, optional

### DIFF
--- a/src/us/kbase/workspace/database/provenance/Organization.java
+++ b/src/us/kbase/workspace/database/provenance/Organization.java
@@ -5,6 +5,9 @@ import static java.util.Objects.requireNonNull;
 import java.util.Objects;
 import java.util.Optional;
 
+/**
+ * Information about an organization.
+ */
 public class Organization {
 	private final String organizationName;
 	private final String organizationID;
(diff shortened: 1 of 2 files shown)


src/us/kbase/workspace/database/provenance/Organization.java
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	/** A builder for an {@link Organization}. */
+	public static class Builder {
+
+		private final String organizationName;
+		private String organizationID = null;
+
+		private Builder(final String organizationName) {
+			this.organizationName = requireNonNull(Common.processString(organizationName), "organizationName");
```
MrCreosote, line 68: I would normally recommend setting a length limit on both the name and ID fields, but

1. The other provenance classes don't
2. There's a 1MB limit on the provenance data structure as a whole when converted to json

src/us/kbase/workspace/database/provenance/Organization.java
```diff
@@ -0,0 +1,92 @@
+package us.kbase.workspace.database.provenance;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class Organization {
```
MrCreosote, line 8: Needs class docs

MrCreosote (changes requested):
Looks great, just needs a class doc blurb

Issue #1, opened 2014-07-31
ws-get bug


olsonanl:
olson@redwood:~$ ws-get --help
Recursive inheritance detected in package 'Types::Serialiser::BooleanBase' at (eval 23) line 76.
BEGIN failed--compilation aborted at /disks/kb/deployment/plbin/ws-get.pl line 12.
Recursive inheritance detected in package 'Types::Serialiser::BooleanBase' at /disks/kb/deployment/plbin/ws-get.pl line 12 during global destruction.

PR #636, opened 2022-10-20
WIP CE-94: Resolvable PID class plus tests

ialarmedalien:
Resolvable PID class:

id: required String, must match `^\w+:\w.+`
description: optional String
relationship_type: optional String


src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -4,19 +4,24 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /* This is an internal only class for shared code between the provenance classes.
  * Tests are all via testing the public provenance class APIs.
  */
 class Common {
-	
-	private Common() {};
+
```
ialarmedalien, line 22: I ran Eclipse's "format" command to fix up the code I'd added, so there are whitespace changes in addition to the new code.

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -47,38 +52,53 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
 			return Common.immutable(list);
 		}
 	}
-	
+
+	static String checkRpid(final String putativeRpid, final String name) {
```
ialarmedalien, line 56: Check RPID format

The rest of the changes to this file are whitespace

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -47,38 +52,53 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
...
+		final String rpid = requireNonNull(processString(putativeRpid), name);
```
MrCreosote, line 57: Take a look at `us.kbase.workspace.database.Util.checkString`

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -47,38 +52,53 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
...
+
+		final Pattern validRpidRegex = Pattern.compile("^\\w+:\\w+");
+		final Matcher m = validRpidRegex.matcher(rpid);
+		if (m.find()) {
+			return rpid;
+		}
+		throw new IllegalArgumentException(String.format(
+				"Illegal ID format for %s: \"%s\"\n" + "ID prefixes should match the pattern \"^\\w+:\\w+\"", name,
```
MrCreosote, line 65: The `+` is just for readability I guess?

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -47,38 +52,53 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
...
+				putativeRpid));
```
MrCreosote, line 66: I would prefer the string on line 1 and the variables on line 2... oh but formatter, nm

src/us/kbase/workspace/database/provenance/ResolvablePID.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * A resolvable permanent identifier for a resource.
+ */
+public class ResolvablePID {
+
+	private final String id;
+	private final String description;
```
MrCreosote, line 12: Is this a description of the ID or the referent of the ID? The docs say it's a description of the ID but that doesn't seem particularly helpful

src/us/kbase/workspace/database/provenance/ResolvablePID.java
```diff
@@ -0,0 +1,118 @@
...
+	 * 
+	 * @return the description, if present.
+	 */
+	public Optional<String> getDescription() {
+		return Optional.ofNullable(description);
+	}
+
+	/**
+	 * Get the relationship between the ID and the resource, for example
+	 * "isSupplementTo" or "isNewVersionOf".
```
MrCreosote, line 41: Hmm, I don't want to open a can of worms, but should this be an Enum or some other type of controlled vocab?

MrCreosote, line 41: Actually I'm a bit confused by this. The class docs say that this class is an ID for a resource, but this seems to imply it's an ID for a resource that is somehow related to a *different* resource.

src/us/kbase/workspace/test/database/provenance/ResolvablePIDTest.java
```diff
@@ -0,0 +1,137 @@
...
+		final ResolvablePID rpid2 = ResolvablePID.getBuilder(RPID_STRING).withDescription(DESC_STRING)
+				.withDescription(null).withRelationshipType(REL_STRING).withRelationshipType(null).build();
+		assertThat(INCORRECT_RPID, rpid2.getId(), is(RPID_STRING));
+		assertThat(INCORRECT_DESC, rpid2.getDescription(), is(Optional.empty()));
+		assertThat(INCORRECT_REL_TYPE, rpid2.getRelationshipType(), is(Optional.empty()));
+	}
+
+	@Test
+	public void buildFailInvalidRPID() throws Exception {
+		try {
```
MrCreosote, line 98: I leave this up to you, but what I would do here is make positive and negative test functions and run a list of positive and negative RPIDs through them with a range of characters and formats

MrCreosote (changes requested):

src/us/kbase/workspace/database/provenance/Common.java
ialarmedalien: yes -- it did span two lines but I didn't notice that the formatter had turned it into a single line.

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -4,19 +4,24 @@
...
 
 /* This is an internal only class for shared code between the provenance classes.
  * Tests are all via testing the public provenance class APIs.
  */
 class Common {
-	
-	private Common() {};
+
+	private Common() {
+	};
```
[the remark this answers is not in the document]
ialarmedalien: If you aren't using the formatter and don't care for it, I won't bother either. The only advantage is that it eliminates differences in formatting preferences (by creating a format that no one likes) and I predict that my preferred code layout is probably going to be different to yours.

src/us/kbase/workspace/database/provenance/ResolvablePID.java
ialarmedalien: Ah, I will clarify that. Thanks!

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
MrCreosote: The purpose of code formatters is to get people to hate the formatter instead of each other which is so much less satisfying

src/us/kbase/workspace/database/provenance/ResolvablePID.java
ialarmedalien: yes, it's coming from a controlled vocab. I'm in two minds about what to do here (and for other fields that would ideally be enums) -- whether it would be better to validate in the credit engine (where I have tools that do all the work for me) and have the WS just save the data, or whether the WS should also validate enums such as these too. Enums are the the parts of the schema that I think are most likely to need editing in the future, and it will be a nuisance if the WS has to be updated each time.

A few other possible ideas I've thought of: 
- build a minimal parser for the schema that could pull in the appropriate enum values for the appropriate classes;
- give the WS a copy of the schema in JSONschema format, validate the incoming data according to the schema (using an existing Java validator), so errors around enums, etc., get caught there and these classes don't need to do much validation beyond type checking; 
- have the WS launch a subprocess that runs the validator (python library);
- extend the existing linkml Java class generator code to be able to create fully-fleshed out Java objects from the schema.

For now, since the credit engine is the only likely source of credit metadata, I feel it's reasonable to go for a simple implementation where the WS doesn't check enums, but I'm not sure that's a good long term solution.

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -47,38 +52,53 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
 			return Common.immutable(list);
 		}
 	}
-	
+
+	static String checkRpid(final String putativeRpid, final String name) {
+		final String rpid = requireNonNull(processString(putativeRpid), name);
+
+		final Pattern validRpidRegex = Pattern.compile("^\\w+:\\w+");
```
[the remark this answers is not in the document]
ialarmedalien: I deliberately left the regex open-ended (no `$`) to allow for any of a myriad of different dbxrefs that I've encountered on my journeys. DOIs can contain all manner of non-alphanumeric characters, so it wouldn't be just accounting for `/` and `.`, and this field also has to accommodate references from other databases that may also play fast and loose with sane standards for PID naming.

This ties in somewhat to the comments below about enums and where validation should occur.

src/us/kbase/workspace/database/provenance/ResolvablePID.java
ialarmedalien: (we did discuss some of this before, with the conclusion being KISS for the time being; however, I'm not sure I mentioned the other strategies that I thought of to deal with the issue)

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
MrCreosote: This is true for the prefix portion of the ID as well?

The reason I was a bit confused is that if there's no anchor, there's no difference between `\w` and `\w+` - both enforce one alphnum character and then there's no constraints on anything after that. The `+` made me think there was a missing anchor.

I'm also wondering, if dbxrefs are so unconstrained, if there should be a constraint on the first character?

src/us/kbase/workspace/database/provenance/Common.java
MrCreosote: I would either revert the formatter change or merge into a single string then

src/us/kbase/workspace/database/provenance/ResolvablePID.java
MrCreosote: Since only admins will be able to add credit info for now, validating outside the WS seems fine, with documentation somewhere in a relevant place in the WS repo warning future coders that they made need to do more validation if the method is opened up.

For the future, I think giving the workspace the JSONschmea seems reasonable, or it probably wouldn't be too hard to make a script that extracts the relationship values from the schema and creates a java Enum (or is that what you mean by #1?).

ialarmedalien: > it probably wouldn't be too hard to make a script that extracts the relationship values from the schema and creates a java Enum (or is that what you mean by https://github.com/kbase/workspace_deluxe/issues/1?).

Yep.

I have been looking at JSONschema tools this morning and there are some Java generators that can produce Java classes (of various different kinds / styles) from a JSONschema, so future enhancements might include adding JSONschema validation to the WS and implementing the Java version of the schema using one of those generators.

src/us/kbase/workspace/database/provenance/Common.java
ialarmedalien: There's also the `checkURL` methods beneath this in this file. What's the logic behind what goes in `Common` vs what goes in `Util`?

MrCreosote: Whether it's only used in provenance classes or not. I imagine some will get moved to other places (I actually would like to change Util to something else at some point as it's not a very descriptive name) at some point.

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
ialarmedalien: Change undone

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -47,38 +52,58 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
...
 	}
-	
+
+	/** Check that a PID string is non-null, has at least one non-whitespace character,
+	 * and conforms to the specified regular expression.
+	 * @param putativePid the string to check.
+	 * @param name the name of the string to use in any error messages.
+	 * @return the trimmed PID.
+	 */
+	public static String checkPid(final String putativePid, final String name)
```
ialarmedalien, line 62: Changed this to `checkPid` and more generally changed `ResolvablePID` to `PermanentID` as PID is a commonly-used acronym (well, Elisha uses it) but RPID isn't.

src/us/kbase/workspace/database/provenance/Common.java
ialarmedalien: Update: undid the format command.

src/us/kbase/workspace/database/provenance/PermanentID.java
```diff
@@ -0,0 +1,120 @@
+package us.kbase.workspace.database.provenance;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Represents a permanent unique identifier for an entity.
+ * Used for the 'related_identifiers' field of a {@link Resource} object.
```
ialarmedalien, line 8: The only place in which this class is used -- 🤞  that makes the confusing documentation a bit less confusing.

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
ialarmedalien: I've changed the regex so that it now just ensures that the first character after the colon is not whitespace. The potential prefixes and prefix formatting should be constrained by the Credit Engine, and the CE should also ensure that the whole thing is in a reasonable form.

src/us/kbase/workspace/database/provenance/Common.java
ialarmedalien: Renamed this function `checkPid` and altered the functionality to be more similar to the `checkString` function in `Util`.

MrCreosote: Oh, I just meant that you could replace `requireNonNull(processString...)` with `checkString`

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
MrCreosote: that's fine, but if there's some general restrictions that are relatively easy to do I don't think there's a reason not to handle them here as well just to be safe

src/us/kbase/workspace/database/provenance/ResolvablePID.java
ialarmedalien: Rewritten the docs so it should be a bit clearer.

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
ialarmedalien: The other restrictions are going to be checked by an external python lib, so not really suitable for adding here.

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/Common.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/Common.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/Common.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/ResolvablePID.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/ResolvablePIDTest.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
MrCreosote: The first part of the PID is now really locked down

PR #639, opened 2022-11-02
CE-93: nitpickery

ialarmedalien:
- new `checkAgainstRegex` method to allow regex matching and optional string replacement
- variable renaming for greater accuracy and consistency


MrCreosote:
Looks like there's some uncovered lines in the new code

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -55,27 +55,62 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
...
 	 *
-	 * @param putativePid the string to check.
-	 * @param name        the name of the string to use in any error messages.
-	 * @param optional    whether or not the field is optional. If true, null is a valid value for the PID.
-	 * @return the trimmed PID.
+	 * @param stringToCheck the string to check.
+	 * @param pattern       the pattern to validate against.
+	 * @param replace       if non-null, the pattern to use for the replaceAll operation.
+	 * @param name          the name of the string to use in any error messages.
+	 * @param optional      whether or not the field is optional. If true, null is a valid value for the field.
```
MrCreosote, line 67: Nit: null or whitespace

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -55,27 +55,62 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
...
+	}
+
+	/**
+	 * Trims leading and trailing whitespace, and then checks that a string is either null, or
+	 * has at least one non-whitespace character and conforms to the specified regular expression.
+	 *
+	 * @param stringToCheck the string to check.
+	 * @param pattern       the pattern to validate against.
+	 * @param name          the name of the string to use in any error messages.
+	 * @param optional      whether or not the field is optional. If true, null is a valid value for the field.
```
MrCreosote, line 95: nit whitespace

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -55,27 +55,62 @@ static <T> List<T> processSimpleList(final List<T> list, final String name) {
...
+		return checkAgainstRegex(stringToCheck, pattern, null, name, optional);
+	}
+
+	/**
+	 * Trims leading and trailing whitespace, and then checks that a putative PID is either null,
+	 * or has at least one non-whitespace character and conforms to the specified regular expression.
+	 *
+	 * @param putativePid the putative PID string.
+	 * @param name        the name of the string to use in any error messages.
+	 * @param optional    whether or not the field is optional. If true, null is a valid value for the field.
```
MrCreosote, line 108: nit whitespace

MrCreosote (changes requested):
The `checkAgainstRegex` methods need their own tests but otherwise LGTM

src/us/kbase/workspace/database/provenance/Common.java
ialarmedalien: > Trims leading and trailing whitespace and then checks that a string is either null, or has at least ...

Whitespace isn't a valid value; null is.

ialarmedalien:
I've got tests for `checkAgainstRegex` incorporated into the next PR.

src/us/kbase/workspace/database/provenance/Common.java
MrCreosote: Unless I'm completely misreading something here, if optional is true both `null` and `\t   \n` will result in this function returning `null`... right?

MrCreosote: Maybe the confusion is over value vs. input, I was thinking input

MrCreosote:
Is there a reason to not have them in this PR? I try to always have tests against any new code in a PR

src/us/kbase/workspace/database/provenance/Common.java
ialarmedalien: Yup, that's it. How about this wording for the param:

```
	 * @param optional      whether or not the field is optional. If false, null,
         *                      empty, or whitespace-only input strings will throw an error.
```

MrCreosote: :+1:

ialarmedalien:
The appropriate tests will be in the CE-94_resolvable_pid PR (#636), which I'll move out of draft mode once it's rebased and updated.

MrCreosote:
In order to get the tests to pass, `next` will need to be changed to `ci` here:
https://github.com/kbase/workspace_deluxe/blob/develop/src/us/kbase/common/test/service/ServiceCheckerTest.java#L20

Probably a better option in any case given the evolution of next and ci

MrCreosote:
also the testPerl test can be dropped given there aren't any Perl SDK core services left

MrCreosote (changes requested):
This looks good other than the fix to get the tests to pass

MrCreosote:
Codecov clearly has the report for the commit but doesn't seem to be able to update the comment:
https://app.codecov.io/github/kbase/workspace_deluxe/commit/7bfd1e7375d8a7c3f77dcbc6d4e77e4f39fb3305

MrCreosote:
Or update the pull request either https://app.codecov.io/github/kbase/workspace_deluxe/pull/639

PR #641, opened 2022-11-03
CE-94 revisited: Adding PID class with tests

ialarmedalien:
PermanentID class:
id: mandatory, must match regex
description: optional, free text
relationshipType: optional, currently free text but may be restricted to a controlled vocab in the future

### DIFF
--- a/src/us/kbase/workspace/database/provenance/PermanentID.java
+++ b/src/us/kbase/workspace/database/provenance/PermanentID.java
@@ -121,7 +121,9 @@ public class PermanentID {
 		/**
 		 * Builds the {@link PermanentID}.
 		 *
-		 * @return the external data.
+		 * @param id  the permanent ID, for example DOI:10.25982/59912.37.
+		 *
+		 * @return the permanent ID.
 		 */
 		public PermanentID build() {
 			return new PermanentID(id, description, relationshipType);
--- a/src/us/kbase/workspace/database/provenance/PermanentID.java
+++ b/src/us/kbase/workspace/database/provenance/PermanentID.java
@@ -1,0 +1,130 @@
+package us.kbase.workspace.database.provenance;
+
+import java.util.Objects;
+import java.util.Optional;
+import us.kbase.workspace.database.Util;
+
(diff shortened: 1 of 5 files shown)


src/us/kbase/workspace/database/provenance/PermanentID.java
```diff
@@ -0,0 +1,130 @@
...
+	 * @return the description, if present.
+	 */
+	public Optional<String> getDescription() {
+		return Optional.ofNullable(description);
+	}
+
+	/**
+	 * Gets the relationship between the ID and some other entity. For
+	 * example, when a {@link PermanentID] class is used to represent
+	 * objects in a {@link Resource}'s 'related_identifiers' field, this
```
MrCreosote, line 45: I presume this'll be added in a later PR

src/us/kbase/workspace/database/provenance/PermanentID.java
```diff
@@ -0,0 +1,130 @@
...
+	public Optional<String> getRelationshipType() {
+		return Optional.ofNullable(relationshipType);
+	}
+
+	/**
+	 * Gets a builder for an {@link PermanentID}.
+	 *
+	 * @return the builder.
+	 */
+	public static Builder getBuilder(final String id) {
```
MrCreosote, line 63: Needs `@param` entry for an id describing the format

MrCreosote (changes requested):
One minor doc update. After that, approval is tentative pending
* tests pass
* coverage is 100% as expected
* the missing coverage in the Common regex matching functions is now covered as discussed

src/us/kbase/workspace/database/provenance/PermanentID.java
ialarmedalien: Yup, correct.

src/us/kbase/workspace/database/provenance/PermanentID.java
ialarmedalien: Done.

PR #642, opened 2022-11-03
Switching tests over to CI whilst next is kaput

ialarmedalien:
As suggested in the PR title.


MrCreosote (approved):
LGTM

And this is forever and ever, no need to switch back to next

MrCreosote:
we should wait for tests to pass before merge though, obs... it'd be nice if we could tell travis which jobs to run first

PR #643, opened 2022-11-08
CE-97: Funding Reference class

ialarmedalien:
Funding reference class:
- funderName: string, required
- funderID: string, PID format, optional
- awardID: string, may or may not be in PID format, optional
- awardTitle: string, optional
- awardURI: URL, optional

This PR needs #639 to be merged first so that the tests will pass.

### DIFF
--- a/src/us/kbase/workspace/database/provenance/FundingReference.java
+++ b/src/us/kbase/workspace/database/provenance/FundingReference.java
@@ -15,19 +15,19 @@ public class FundingReference {
 	private final String funderName;
 	private final String awardID;
 	private final String awardTitle;
-	private final URL awardURI;
+	private final URL awardURL;
 
 	private FundingReference(
 			final String funderID,
 			final String funderName,
 			final String awardID,
 			final String awardTitle,
-			final URL awardURI) {
+			final URL awardURL) {
 		this.funderID = funderID;
 		this.funderName = funderName;
 		this.awardID = awardID;
 		this.awardTitle = awardTitle;
-		this.awardURI = awardURI;
+		this.awardURL = awardURL;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/us/kbase/workspace/test/database/provenance/FundingReferenceTest.java
```diff
@@ -0,0 +1,183 @@
...
+	@Test
+	public void buildFailInvalidPIDs() throws Exception {
+		for (String invalidPid : INVALID_PID_LIST) {
+			try {
+				FundingReference.getBuilder(FUNDER_NAME).withFunderID(invalidPid).build();
+				fail(EXP_EXC);
+			} catch (Exception got) {
+				TestCommon.assertExceptionCorrect(got, new IllegalArgumentException(
+					"Illegal format for funderID: \"" + invalidPid + "\"\n" +
+					"It should match the pattern \"^([a-zA-Z0-9][a-zA-Z0-9\\.]+)\\s*:\\s*(\\S.+)$\""));
```
ialarmedalien, line 152: this test currently fails; it will pass once #639 is merged.

src/us/kbase/workspace/database/provenance/FundingReference.java
```diff
@@ -0,0 +1,194 @@
...
+	public Optional<String> getAwardTitle() {
+		return Optional.ofNullable(awardTitle);
+	}
+
+	/**
+	 * Gets the URI of the award.
+	 *
+	 * @return the award URI, if present.
+	 */
+	public Optional<URL> getAwardURI() {
```
MrCreosote, line 76: Hmm, method says it's returning a URI but actually returns a URL

MrCreosote (changes requested):
Looks good other than inconsistent names / returns for the URL/URI methods

MrCreosote (approved):

PR #645, opened 2022-11-11
CE-96: title class

ialarmedalien:
Fields:
`titleString` - required
`titleType` - optional, enum
`titleLanguage` - optional

### DIFF
--- a/src/us/kbase/workspace/database/provenance/Title.java
+++ b/src/us/kbase/workspace/database/provenance/Title.java
@@ -1,24 +1,25 @@
 package us.kbase.workspace.database.provenance;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.HashMap;
 import java.util.Objects;
 import java.util.Optional;
 import us.kbase.workspace.database.Util;
+import java.util.Locale;
+import java.util.IllformedLocaleException;
 
 /**
  * Represents a title or name of a resource.
  */
 public class Title {
 
-	private final String titleString;
-	private final TitleType titleType;
-	private final String titleLanguage;
-
 	/**
 	 * An enum representing types of title.
 	 */
-	public static enum TitleType {
+	public enum TitleType {
+		TITLE,
 		SUBTITLE,
 		ALTERNATIVE_TITLE,
 		TRANSLATED_TITLE,
@@ -26,24 +27,34 @@ public class Title {
 
 		private static final Map<String, TitleType> STRING_TO_TITLE_TYPE_MAP = new HashMap<>();
 		static {
-			for (final TitleType tt: TitleType.values()) {
+			for (final TitleType tt : TitleType.values()) {
 				STRING_TO_TITLE_TYPE_MAP.put(tt.name().toLowerCase(), tt);
 			}
 		}
 
-		/** Gets a title type based on a supplied string.
-		 * @param input a string representing a title type.
+		/**
+		 * Gets a title type based on a supplied string.
+		 *
+		 * @param titleType
+		 *                a string representing a title type.
 		 * @return a TitleType.
-		 * @throws IllegalArgumentException if there is no title type related to the input string.
+		 * @throws IllegalArgumentException
+		 *                 if there is no title type related to the input string.
 		 */
-		public static TitleType getTitleType(final String input) {
-			final String lowercaseInput = Util.checkString(input, "input").toLowerCase();
+		public static TitleType getTitleType(final String titleType) {
+			final String lowercaseInput = Util.checkString(titleType, "titleType").toLowerCase();
 			if (STRING_TO_TITLE_TYPE_MAP.containsKey(lowercaseInput)) {
 				return STRING_TO_TITLE_TYPE_MAP.get(lowercaseInput);
 			}
-			throw new IllegalArgumentException("Invalid titleType: " + input);
+			throw new IllegalArgumentException("Invalid titleType: " + titleType);
 		}
 	}
+
+	private static final TitleType DEFAULT_TITLE_TYPE = TitleType.TITLE;
+
+	private final String titleString;
+	private final TitleType titleType;
+	private final String titleLanguage;
 
 	private Title(
 			final String titleString,
@@ -66,14 +77,15 @@ public class Title {
 	/**
 	 * Gets the type of title being represented.
 	 *
-	 * @return the titleType, if present.
+	 * @return the titleType; a null value returns the default titleType
+	 *         TitleType.TITLE.
 	 */
-	public Optional<TitleType> getTitleType() {
-		return Optional.ofNullable(titleType);
+	public TitleType getTitleType() {
+		return titleType;
 	}
 
 	/**
-	 * Gets the language in which the title is written.
+	 * Gets the language in which the title is written, in IETF BCP-47-compliant syntax.
 	 *
 	 * @return the titleLanguage, if present.
 	 */
@@ -115,16 +127,22 @@ public class Title {
 		private String titleString;
 		private TitleType titleType = null;
 		private String titleLanguage = null;
+		private List<String> errorList = new ArrayList<>();
 
 		private Builder(final String titleString) {
-			this.titleString = Util.checkString(titleString, "titleString");
+			try {
+				this.titleString = Util.checkString(titleString, "titleString");
+			} catch (Exception e) {
+				this.errorList.add(e.getMessage());
+			}
 		}
 
 		/**
-		 * Sets the language of title being represented, for example en-GB.
+		 * Sets the language of title being represented, for example en_GB.
 		 *
-		 * @param titleLanguage the language. Null or the empty string removes any
-		 *                      current resource in the builder.
+		 * @param titleLanguage
+		 *                the language. Null or the empty string removes any
+		 *                current value in the builder.
 		 * @return this builder.
 		 */
 		public Builder withTitleLanguage(final String titleLanguage) {
@@ -135,23 +153,32 @@ public class Title {
 		/**
 		 * Sets the type of the title being represented, for example translated_title.
 		 *
-		 * @param titleType the titleType as a string. Null or the empty string removes any
-		 *                  current titleType in the builder.
+		 * @param titleType
+		 *                the titleType as a string. Null or the empty string resets
+		 *                the current titleType in the builder.
 		 * @return this builder.
 		 */
 		public Builder withTitleType(final String titleType) {
 			final String protoTitleType = Common.processString(titleType);
-			this.titleType = protoTitleType == null
-				? null
-				: TitleType.getTitleType(protoTitleType);
+			if (protoTitleType == null) {
+				this.titleType = null;
+			} else {
+				try {
+					this.titleType = TitleType.getTitleType(protoTitleType);
+				} catch (IllegalArgumentException e) {
+					this.errorList.add(e.getMessage());
+				}
+			}
+
 			return this;
 		}
 
 		/**
 		 * Sets the type of the title being represented, for example translated_title.
 		 *
-		 * @param titleType the titleType as an enum. Null removes any
-		 *                  current titleType in the builder.
+		 * @param titleType
+		 *                the titleType as an enum. Null resets any
+		 *                current titleType in the builder.
 		 * @return this builder.
 		 */
 		public Builder withTitleType(final TitleType titleType) {
@@ -162,10 +189,41 @@ public class Title {
 		/**
 		 * Build the {@link Title}.
 		 *
+		 * The titleLanguage value is checked for conformance to IETF BCP-47 syntax
+		 * rules; note that Java does not check the values themselves, but rather the
+		 * syntax of the value.
+		 *
+		 * If the titleType is null, it is set to the default DEFAULT_TITLE_TYPE.
+		 *
 		 * @return the title object.
 		 */
 		public Title build() {
-			return new Title(titleString, titleType, titleLanguage);
+
+			// check and reformat the titleLanguage
+			if (titleLanguage != null) {
+				titleLanguage = titleLanguage.replace("_", "-");
+				Locale.Builder localeBuilder = null;
+				try {
+					localeBuilder = new Locale.Builder().setLanguageTag(titleLanguage);
+				} catch (IllformedLocaleException e) {
+					errorList.add("titleLanguage error: " + e.getMessage());
+				}
+				if (localeBuilder != null) {
+					final Locale l = localeBuilder.build();
+					titleLanguage = l.toLanguageTag();
+				}
+			}
+
+			if (titleType == null) {
+				titleType = DEFAULT_TITLE_TYPE;
+			}
+
+			if (errorList.isEmpty()) {
+				return new Title(titleString, titleType, titleLanguage);
+			}
+
+			throw new IllegalArgumentException("Errors in Title construction:\n" +
+					String.join("\n", errorList));
 		}
 	}
 }
--- a/src/us/kbase/workspace/database/provenance/Title.java
+++ b/src/us/kbase/workspace/database/provenance/Title.java
@@ -18,7 +18,7 @@ public class Title {
 	/**
 	 * An enum representing types of title.
 	 */
-	public enum TitleType {
+	public static enum TitleType {
 		SUBTITLE,
 		ALTERNATIVE_TITLE,
 		TRANSLATED_TITLE,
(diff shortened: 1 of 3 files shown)


src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,171 @@
...
+import us.kbase.workspace.database.Util;
+
+/**
+ * Represents a title or name of a resource.
+ */
+public class Title {
+
+	private final String titleString;
+	private final TitleType titleType;
+	private final String titleLanguage;
```
MrCreosote, line 16: Presumably this ideally would be an enum but that seems like a mess in the making. It'd be nice if there were some way of preventing misspellings / alternate spellings of common languages

src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * An enum representing types of title.
+	 */
+	public enum TitleType {
```
MrCreosote, line 21:
```suggestion
	public static enum TitleType {
```
No reason to bind an enum instance to its enclosing Title instance AFAICT

src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,171 @@
...
+		SUBTITLE,
```
MrCreosote, line 22: Shouldn't just TITLE be one of the options? Or does the lack of a title type imply it's a regular title?

src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,171 @@
...
+	public Optional<TitleType> getTitleType() {
+		return Optional.ofNullable(titleType);
+	}
+
+	/**
+	 * Gets the language in which the title is written.
+	 *
+	 * @return the titleLanguage, if present.
+	 */
+	public Optional<String> getTitleLanguage() {
```
MrCreosote, line 80: I feel like this shouldn't be optional but I don't understand the context

src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,171 @@
...
+		private String titleString;
+		private TitleType titleType = null;
+		private String titleLanguage = null;
+
+		private Builder(final String titleString) {
+			this.titleString = Util.checkString(titleString, "titleString");
+		}
+
+		/**
+		 * Sets the language of title being represented, for example en-GB.
```
MrCreosote, line 124: if you're using IETF format for language identifiers it'd be great to check that at least

MrCreosote (commented):
Looks good, just some questions and possible improvements for dicussion

src/us/kbase/workspace/database/provenance/Title.java
ialarmedalien: yes -- titleType is only used if it's not a plain old title.

src/us/kbase/workspace/database/provenance/Title.java
ialarmedalien: With both this and the titleType, I was weighing up whether to have default values (`en` in this case, plain old title in the case of `titleType`) or whether to leave them null if the value is unknown or unrecorded. I went for the conservative approach, leaving them null by default, as there's less likelihood that a title would have an incorrect titleType or titleLanguage, than if defaults were assigned. This is what DataCite, from whom we will be sourcing a fair bit of our data, do.

src/us/kbase/workspace/database/provenance/Title.java
ialarmedalien: I plan to retrofit the language codes as an enum -- I haven't yet found a good source but it's on the to-do list.

src/us/kbase/workspace/database/provenance/Title.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/Title.java
MrCreosote: Can we document that? I don't think the documentation makes that clear

MrCreosote: Alternatively, add TITLE and require the enum... that might be simpler and more explicit

src/us/kbase/workspace/database/provenance/Title.java
MrCreosote: Fair enough.

src/us/kbase/workspace/database/provenance/Title.java
MrCreosote: https://stackoverflow.com/questions/60795579/validating-bcp-47-language-tag-using-java-apis didn't read closely but might do what you want

src/us/kbase/workspace/database/provenance/Title.java
MrCreosote: And potentially add UNKNOWN as a default if there's no info

src/us/kbase/workspace/database/provenance/Title.java
MrCreosote: Also https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html

MrCreosote: https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#forLanguageTag-java.lang.String-

ialarmedalien: Great, thanks for those links!

src/us/kbase/workspace/database/provenance/Title.java
ialarmedalien: The info about the "missing" default is in the docs for the schema but not here, so that can easily be corrected. I'll do the same with the assumed language.

src/us/kbase/workspace/database/provenance/Title.java
ialarmedalien: After much research: Java does not have a BCP-47 validator; I can check that the input has the appropriate syntax but there's no built-in method for asserting that the values are valid language/region/script/extlang tags. I would need to pull in the [source files](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) and build my own parser/validator. If I'm going to do that, I think I'd prefer to either use an existing lib (there are some py libs available) or parse/convert the file to something like JSONschema, which can be used elsewhere, not just in these Java classes.

MrCreosote: `Locale` doesn't do what you want?

ialarmedalien: No, it just checks that the syntax is OK. I've used its capabilities such as they are, so at least the values can be checked for conformance with BCP-47 syntax rules.

src/us/kbase/workspace/database/provenance/Title.java
ialarmedalien: Update: added the default `TITLE` value to the enum.

src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,229 @@
...
+				}
+				if (localeBuilder != null) {
+					final Locale l = localeBuilder.build();
+					titleLanguage = l.toLanguageTag();
+				}
+			}
+
+			if (titleType == null) {
+				titleType = DEFAULT_TITLE_TYPE;
+			}
```
ialarmedalien, line 219: use the default value for the titleType if nothing is set

src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,229 @@
...
+				titleLanguage = titleLanguage.replace("_", "-");
+				Locale.Builder localeBuilder = null;
+				try {
+					localeBuilder = new Locale.Builder().setLanguageTag(titleLanguage);
+				} catch (IllformedLocaleException e) {
+					errorList.add("titleLanguage error: " + e.getMessage());
+				}
+				if (localeBuilder != null) {
+					final Locale l = localeBuilder.build();
+					titleLanguage = l.toLanguageTag();
```
ialarmedalien, line 213: ensures formatting is consistent

src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,229 @@
...
+		 * @return the title object.
+		 */
+		public Title build() {
+
+			// check and reformat the titleLanguage
+			if (titleLanguage != null) {
+				titleLanguage = titleLanguage.replace("_", "-");
+				Locale.Builder localeBuilder = null;
+				try {
+					localeBuilder = new Locale.Builder().setLanguageTag(titleLanguage);
```
ialarmedalien, line 207: Try using the supplied language tag to build a locale. Java checks the syntax of the tag but doesn't check the values themselves.

src/us/kbase/workspace/database/provenance/Title.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/Title.java
MrCreosote: Well, that's poopy

src/us/kbase/workspace/test/database/provenance/TitleTest.java
```diff
@@ -0,0 +1,225 @@
...
+				.withTitleType(TYPE_STRING_UNTRIMMED)
+				.withTitleLanguage(LANG_STRING_UNTRIMMED).build();
+		assertTitleFields(title, TitleType.TRANSLATED_TITLE, LANG_STRING);
+	}
+
+	@Test
+	public void buildTransformTitleLanguage() throws Exception {
+		// these tags are valid BCP-47 syntax, even if the values themselves are
+		// nonsense.
+		final String[][] langPairs = {
```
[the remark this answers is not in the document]
ialarmedalien: It started off as an array but turned into an array of arrays after I decided to add in a formatting element (e.g. input: `ZH_HANT_CN` => output: `zh-Hant-CN`).

I also dislike the standard Map because there's no graceful way to initialise it with values in Java 8. Even the ImmutableMap from Guava maxes out at five key-value pairs (which is sensible, given the syntax, but why not use a different syntax that's less prone to confusion?). Grumble grumble.

src/us/kbase/workspace/database/provenance/Title.java
```diff
@@ -0,0 +1,229 @@
...
+		 *
+		 * If the titleType is null, it is set to the default DEFAULT_TITLE_TYPE.
+		 *
+		 * @return the title object.
+		 */
+		public Title build() {
+
+			// check and reformat the titleLanguage
+			if (titleLanguage != null) {
+				titleLanguage = titleLanguage.replace("_", "-");
```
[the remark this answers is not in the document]
ialarmedalien: If you have some examples, that'd be great.

src/us/kbase/workspace/test/database/provenance/TitleTest.java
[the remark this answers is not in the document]
MrCreosote: I do this: https://github.com/kbase/workspace_deluxe/blob/3d716809d838fd46d0d3170cfea59a647bf0a1b6/src/us/kbase/workspace/test/kbase/KBaseWorkspaceConfigTest.java#L441-L449

... which is still clunkier than a literal, but better than Guava or regular maps IMO

src/us/kbase/workspace/database/provenance/Title.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/TitleTest.java
```diff
@@ -0,0 +1,225 @@
...
+	private void assertTitleFields(final Title title, TitleType titleType, String titleLanguage) {
+		assertThat(INCORRECT_TITLE, title.getTitleString(), is(TITLE_STRING));
+		assertThat(INCORRECT_TYPE, title.getTitleType(), is(
+				titleType == null ? TitleType.TITLE : titleType));
+		assertThat(INCORRECT_LANG, title.getTitleLanguage(), is(
+				titleLanguage == null ? ES : opt(titleLanguage)));
+	}
+
+	private void assertTitleFields(final Title title) {
+		assertTitleFields(title, null, null);
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/TitleTest.java
```diff
@@ -0,0 +1,225 @@
...
+	static final String TITLE_STRING = "A Series of Unfortunate Elephants";
+	static final String TITLE_STRING_UNTRIMMED = "\t\t\f\t\n\rA Series of Unfortunate Elephants\n\n\n   ";
+
+	static final String TYPE_STRING = "translated_title";
+	static final String TYPE_STRING_UNTRIMMED = "  \n translated_title   \t\n  ";
+
+	static final String LANG_STRING = "de";
+	static final String LANG_STRING_UNTRIMMED = "\nde\n";
+
+	static final TitleType DEFAULT_TITLE_TYPE = TitleType.TITLE;
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/TitleTest.java
```diff
@@ -0,0 +1,225 @@
...
+	public void equals() throws Exception {
+		EqualsVerifier.forClass(Title.class).usingGetClass().verify();
+	}
+
+	private void assertTitleFields(final Title title, TitleType titleType, String titleLanguage) {
+		assertThat(INCORRECT_TITLE, title.getTitleString(), is(TITLE_STRING));
+		assertThat(INCORRECT_TYPE, title.getTitleType(), is(
+				titleType == null ? TitleType.TITLE : titleType));
+		assertThat(INCORRECT_LANG, title.getTitleLanguage(), is(
+				titleLanguage == null ? ES : opt(titleLanguage)));
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/TitleTest.java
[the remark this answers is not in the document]
MrCreosote: also java 9 has https://docs.oracle.com/javase/9/docs/api/java/util/Map.html#of-K-V-K-V-K-V-K-V-K-V-K-V-K-V-K-V-K-V-K-V-

Unfortunately getting off 8 is going to be a bear. I've started gradelizing stuff in my spare time to prep for that but it's taking  a while

src/us/kbase/workspace/database/provenance/Title.java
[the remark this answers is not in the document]
ialarmedalien: Added to the test set, although it gets converted back to `zh-Hant` when running `.toLanguageTag()` on the locale.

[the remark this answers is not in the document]
MrCreosote: I was more wondering if it would correctly handle something like zh__Hant as input - but as I say, that's pretty edgy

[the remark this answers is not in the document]
ialarmedalien: The parser seems to be very simple -- if there's a trailing `-`, it says there's a missing subtag. Similarly, `--` gets reported as a missing subtag at position 1.

src/us/kbase/workspace/test/database/provenance/TitleTest.java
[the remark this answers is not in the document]
ialarmedalien: Yes, I suspected that would be the case! At least Java maintains LTS for old versions so keeping a codebase at v8 is a viable (if not always optimal) option.

src/us/kbase/workspace/database/provenance/Title.java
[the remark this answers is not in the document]
MrCreosote: So the question is whether you want to manually munge the `__` to `-` or not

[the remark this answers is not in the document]
ialarmedalien: I'll leave as-is here -- it looks like there are some python libraries that will check the values as well as the formatting of bcp-47 tags, so ideally there'll be a validation step in the credit engine code that will catch errors before data gets sent off to the workspace.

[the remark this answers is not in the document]
MrCreosote: :+1:

MrCreosote (approved):
LGTM assuming tests pass

PR #647, opened 2022-11-17
Switch tests from Travis to GHA

### LINKED ISSUE #410: Switch builds from travis to github actions
Might need to update how image is built and pushed as well if that's happening in travis

MrCreosote:
Closes #410


.github/workflows/test.yml
```diff
@@ -0,0 +1,140 @@
...
+          - java: '8'
+            mongo: 'mongodb-linux-x86_64-3.6.23'
+            minio: '2019-05-23T00-29-34Z'
+            wired_tiger: 'false'
+            ant_test: 'test_quick_coverage'
+          - java: '11'
+            mongo: 'mongodb-linux-x86_64-3.6.23'
+            minio: '2019-05-23T00-29-34Z'
+            wired_tiger: 'true'
+            ant_test: 'test_quick_coverage'
```
ialarmedalien, line 35: unlike the Travis tests, you just want to run java 8 / wired_tiger: false, and java 11 / wired_tiger: true; correct?

MrCreosote: yeah, the only reason I did it that way before was because I don't think there was a way to matrix in the java versions. Now you can, so the number of tests can be halved.

*Maybe* there's some crazy interaction between the Mongo storage engine and the java version, but that seems way too unlikely to double our test compute just in case

MrCreosote:
Ugh, I've got to get around to fixing that mongo cleanup bug at some point

ialarmedalien (approved):

PR #652, opened 2022-11-29
Remove trailing whitespace

ialarmedalien:
The ArgUtils file will also be edited as part of this project, so doing the trailing whitespace removal here, too.


MrCreosote (approved):
LGTM

PR #658, opened 2023-01-05
First draft of workspace spec

ialarmedalien:
A start on writing up the workspace spec for the new classes. I anticipate this will require some back and forth, and once the wording looks OK, I will sort out the layout and add in the autogenerated classes.

The specs cover three classes that have not yet been added under `us.kbase.workspace.database.provenance`; this is deliberate as they are more complex and I want to look at how they will be created/manipulated in `ArgUtils` when I add them.

### DIFF
--- a/workspace.spec
+++ b/workspace.spec
@@ -585,9 +585,9 @@ module Workspace {
 		When requesting credit metadata, the last (i.e. most recent) entry in the list is returned.
 
 		In the following entries, "resource" is used to refer to the workspace object that the CM pertains to.
-	*/
 
-	/* EventDate
+
+	EventDate
 
 		Represents an event in the lifecycle of a resource and the date it occurred on.
 
@@ -597,18 +597,24 @@ module Workspace {
 
 		string date - the date associated with the event. The date may be in the format
 			YYYY, YYYY-MM, or YYYY-MM-DD.
-		string event - the event that occurred; events must be from the following list:
-			"accepted"
-			"available"
-			"copyrighted"
-			"collected"
-			"created"
-			"issued"
-			"submitted"
-			"updated"
-			"valid"
-			"withdrawn"
-			"other"
+			Examples:
+				- 2001
+				- 2021-05
+				- 1998-02-15
+
+		string event - the event that occurred.
+			Valid 'event' values:
+				- accepted
+				- available
+				- copyrighted
+				- collected
+				- created
+				- issued
+				- submitted
+				- updated
+				- valid
+				- withdrawn
+				- other
 	*/
 	typedef structure {
 		string date;
@@ -630,10 +636,28 @@ module Workspace {
 
 		string funder_id (optional) - persistent unique identifier for the funder in the format
 			<database name>:<identifier within database>
+			Examples:
+				- ROR:04xm1d337
+				- ISNI:0000000405337147
+
 		string funder_name - common name for the funder
+			Examples:
+				- Joint Genome Institute
+				- National Science Foundation
+
 		string award_id (optional) - code for the award, assigned by the funder
+			Examples:
+				- 1296
+				- CBET-0756451
+
 		string award_title (optional) - title for the award
+			Examples:
+				- Metagenomic analysis of the rhizosphere of three biofuel crops at the KBS intensive site
+
 		string award_url (optional) - URL for the award
+			Examples:
+				- https://genome.jgi.doe.gov/portal/Metanaintenssite/Metanaintenssite.info.html
+
 	*/
 	typedef structure {
 		string funder_id;
@@ -659,11 +683,16 @@ module Workspace {
 		'organization_name' is required; 'organization_id' is optional.
 
 		string organization_name - common name; use the name recommended by ROR if possible.
-		string organization_id (optional) - persistent unique identifier for the organization
-		in the format
-			<database name>:<identifier within database>
+			Examples:
+				- KBase
+				- United States Department of Energy
 
-		e.g. ROR:01znn6x10
+		string organization_id (optional) - persistent unique identifier for the organization
+		in the format <database name>:<identifier within database>
+			Examples:
+				- ROR:01bj3aw27
+				- ISNI:0000000123423717
+				- CrossrefFunder:100000015
 	*/
 	typedef structure {
 		string organization_name;
@@ -676,88 +705,92 @@ module Workspace {
 
 		The 'id' field is required; all other fields are optional.
 
-		The values in the 'relationship_type' field come from controlled vocabularies maintained by Datacite and CrossRef. See the documentation links below for more details.
+		The values in the 'relationship_type' field come from controlled vocabularies maintained by DataCite and Crossref. See the documentation links below for more details.
 
 		DataCite relation types: https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#12b-relationtype
 
-		CrossRef relation types: https://www.crossref.org/documentation/schema-library/markup-guide-metadata-segments/relationships/
+		Crossref relation types: https://www.crossref.org/documentation/schema-library/markup-guide-metadata-segments/relationships/
 
 		string id - persistent unique ID for an entity. Should be in the format
 			<database name>:<identifier within database>
-			e.g.	DOI:10.46936/10.25585/60000745
-				GO:0005456
-				HGNC:7470
+			Examples:
+				- DOI:10.46936/10.25585/60000745
+				- GO:0005456
+				- HGNC:7470
 		string description (optional) - description of that entity
+			Examples:
+				- Amaranthus hypochondriacus genome
+
 		string relationship_type (optional) - the relationship between the ID and some
 			other entity.
 			For example, when a PermanentID class is used to represent objects in the
 			CreditMetadata field 'related_identifiers', the 'relationship_type' field
 			captures the relationship between the CreditMetadata and this ID.
 
-			Valid relationship_type values:
-				"Datacite:Cites"
-				"Datacite:Compiles"
-				"Datacite:Continues"
-				"Datacite:Describes"
-				"Datacite:Documents"
-				"Datacite:HasMetadata"
-				"Datacite:HasPart"
-				"Datacite:HasVersion"
-				"Datacite:IsCitedBy"
-				"Datacite:isCompiledBy"
-				"Datacite:IsContinuedBy"
-				"Datacite:IsDerivedFrom"
-				"Datacite:IsDescribedBy"
-				"Datacite:IsDocumentedBy"
-				"Datacite:IsIdenticalTo"
-				"Datacite:IsMetadataFor"
-				"Datacite:IsNewVersionOf"
-				"Datacite:IsOriginalFormOf"
-				"Datacite:IsPartOf"
-				"Datacite:IsPreviousVersionOf"
-				"Datacite:IsPublishedIn"
-				"Datacite:IsReferencedBy"
-				"Datacite:IsRequiredBy"
-				"Datacite:IsReviewedBy"
-				"Datacite:IsSourceOf"
-				"Datacite:IsSupplementTo"
-				"Datacite:IsSupplementedBy"
-				"Datacite:IsVariantFormOf"
-				"Datacite:IsVersionOf"
-				"Datacite:Obsoletes"
-				"Datacite:References"
-				"Datacite:Requires"
-				"Datacite:Reviews"
-				"CrossRef:BasedOnData"
-				"CrossRef:Finances"
-				"CrossRef:HasComment"
-				"CrossRef:HasDerivation"
-				"CrossRef:HasExpression"
-				"CrossRef:HasFormat"
-				"CrossRef:HasManifestation"
-				"CrossRef:HasManuscript"
-				"CrossRef:HasPreprint"
-				"CrossRef:HasRelatedMaterial"
-				"CrossRef:HasReply"
-				"CrossRef:HasReview"
-				"CrossRef:HasTranslation"
-				"CrossRef:IsBasedOn"
-				"CrossRef:IsBasisFor"
-				"CrossRef:IsCommentOn"
-				"CrossRef:IsDataBasisFor"
-				"CrossRef:IsExpressionOf"
-				"CrossRef:IsFinancedBy"
-				"CrossRef:IsFormatOf"
-				"CrossRef:IsManifestationOf"
-				"CrossRef:IsManuscriptOf"
-				"CrossRef:IsPreprintOf"
-				"CrossRef:IsRelatedMaterial"
-				"CrossRef:IsReplacedBy"
-				"CrossRef:IsReplyTo"
-				"CrossRef:IsReviewOf"
-				"CrossRef:IsSameAs"
-				"CrossRef:IsTranslationOf"
-				"CrossRef:Replaces"
+			Valid 'relationship_type' values:
+				- DataCite:Cites
+				- DataCite:Compiles
+				- DataCite:Continues
+				- DataCite:Describes
+				- DataCite:Documents
+				- DataCite:HasMetadata
+				- DataCite:HasPart
+				- DataCite:HasVersion
+				- DataCite:IsCitedBy
+				- DataCite:isCompiledBy
+				- DataCite:IsContinuedBy
+				- DataCite:IsDerivedFrom
+				- DataCite:IsDescribedBy
+				- DataCite:IsDocumentedBy
+				- DataCite:IsIdenticalTo
+				- DataCite:IsMetadataFor
+				- DataCite:IsNewVersionOf
+				- DataCite:IsOriginalFormOf
+				- DataCite:IsPartOf
+				- DataCite:IsPreviousVersionOf
+				- DataCite:IsPublishedIn
+				- DataCite:IsReferencedBy
+				- DataCite:IsRequiredBy
+				- DataCite:IsReviewedBy
+				- DataCite:IsSourceOf
+				- DataCite:IsSupplementTo
+				- DataCite:IsSupplementedBy
+				- DataCite:IsVariantFormOf
+				- DataCite:IsVersionOf
+				- DataCite:Obsoletes
+				- DataCite:References
+				- DataCite:Requires
+				- DataCite:Reviews
+				- Crossref:BasedOnData
+				- Crossref:Finances
+				- Crossref:HasComment
+				- Crossref:HasDerivation
+				- Crossref:HasExpression
+				- Crossref:HasFormat
+				- Crossref:HasManifestation
+				- Crossref:HasManuscript
+				- Crossref:HasPreprint
+				- Crossref:HasRelatedMaterial
+				- Crossref:HasReply
+				- Crossref:HasReview
+				- Crossref:HasTranslation
+				- Crossref:IsBasedOn
+				- Crossref:IsBasisFor
+				- Crossref:IsCommentOn
+				- Crossref:IsDataBasisFor
+				- Crossref:IsExpressionOf
+				- Crossref:IsFinancedBy
+				- Crossref:IsFormatOf
+				- Crossref:IsManifestationOf
+				- Crossref:IsManuscriptOf
+				- Crossref:IsPreprintOf
+				- Crossref:IsRelatedMaterial
+				- Crossref:IsReplacedBy
+				- Crossref:IsReplyTo
+				- Crossref:IsReviewOf
+				- Crossref:IsSameAs
+				- Crossref:IsTranslationOf
+				- Crossref:Replaces
 	*/
 	typedef structure {
 		string id;
@@ -772,15 +805,28 @@ module Workspace {
 		defaults to 'title'. If the 'title_type' is set to 'translated_title', the appropriate
 		BCP-47 tag must be supplied in the 'title_language' field.
 
+		Note that the workspace checks that the title_language field adheres to IETF BCP-47 syntax rules, but it does not check the validity of the tag.
+
 		string title_string - the resource title
+			Examples:
+				- Amaranthus hypochondriacus
+				- Viral Annotation Pipeline in KBase
+
 		string title_language (optional) - language that the title is in, as a IETF BCP-47 tag.
-			Examples: 'fr'; 'jp-JP'; 'zh-Hant-CN'; 'en-Latn-GB'; 'mn-Cyrl'.
-		string title_type (optional) - the type of title described; must be one of the following:
-			"title" (default)
-			"subtitle"
-			"alternative_title"
-			"translated_title"
-			"other"
+			Examples:
+				- fr
+				- jp-JP
+				- zh-Hant-CN
+				- en-Latn-GB
+				- mn-Cyrl
+
+		string title_type (optional) - the type of title described.
+			Valid 'title_type' values:
+				- title
+				- subtitle
+				- alternative_title
+				- translated_title
+				- other
 	*/
 	typedef structure {
 		string title_string;
@@ -800,67 +846,89 @@ module Workspace {
 		the 'name' field must be populated.
 
 		An individual can be represented by either 'given_name' and 'family_name'; a 'name'
-		(if the name is not in the binomial format); or a `contributor_id`, such as an
+		(if the name is not in the binomial format); or a 'contributor_id', such as an
 		ORCID.
 
-		The 'contributor_role' field takes values from the Datacite and CRediT contributor
+		The 'contributor_role' field takes values from the DataCite and CRediT contributor
 		roles vocabularies. For more information on these resources and choosing the
 		appropriate roles, please see the following links:
 
-		Datacite contributor roles: https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#7a-contributortype
+		DataCite contributor roles: https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#7a-contributortype
 
 		CRediT contributor role taxonomy: https://credit.niso.org
 
 		string contributor_type - must be either 'person' or 'organization'
+			Valid 'contributor_type' values:
+				- person
+				- organization
+
 		string contributor_id (optional) - persistent unique identifier for the contributor;
-			e.g. an ORCID for an individual, or a ROR ID for an organization.
+			this might be an ORCID for an individual, or a ROR ID for an organization.
+			Examples:
+				- ORCID:0000000010001234
+				- ROR:01bj3aw27
+
 		string given_name (optional) - first name, for individuals whose names can be split
 			into given and surnames
+			Examples:
+				- Dolly
+				- Marionetta
+
 		string family_name (optional) - family name, for individuals whose names can be
 			split into given and surnames.
+			Examples:
+				- Parton
+				- de la Carte-Postale
+
 		string name (optional) - contributor name, in cases where the contributor is an
 			organization or a person whose name cannot easily fit into the given/family
 			name structure.
+			Examples:
+				- Madonna
+				- Monarch Initiative
+				- The Incredible Hulk
+
 		list<Organization> affiliations (optional) - list of organizations with which the
 			contributor is affiliated
+
 		list<string> contributor_roles (optional) - list of roles played by the contributor
 			when working on the resource.
-			Valid contributor_role values:
-			"Datacite:ContactPerson"
-			"Datacite:DataCollector"
-			"Datacite:DataCurator"
-			"Datacite:DataManager"
-			"Datacite:Distributor"
-			"Datacite:Editor"
-			"Datacite:HostingInstitution"
-			"Datacite:Producer"
-			"Datacite:ProjectLeader"
-			"Datacite:ProjectManager"
-			"Datacite:ProjectMember"
-			"Datacite:RegistrationAgency"
-			"Datacite:RegistrationAuthority"
-			"Datacite:RelatedPerson"
-			"Datacite:Researcher"
-			"Datacite:ResearchGroup"
-			"Datacite:RightsHolder"
-			"Datacite:Sponsor"
-			"Datacite:Supervisor"
-			"Datacite:WorkPackageLeader"
-			"Datacite:Other"
-			"CRediT:conceptualization"
-			"CRediT:data-curation"
-			"CRediT:formal-analysis"
-			"CRediT:funding-acquisition"
-			"CRediT:investigation"
-			"CRediT:methodology"
-			"CRediT:project-administration"
-			"CRediT:resources"
-			"CRediT:software"
-			"CRediT:supervision"
-			"CRediT:validation"
-			"CRediT:visualization"
-			"CRediT:writing-original-draft"
-			"CRediT:writing-review-editing"
+			Valid 'contributor_role' values:
+				- DataCite:ContactPerson
+				- DataCite:DataCollector
+				- DataCite:DataCurator
+				- DataCite:DataManager
+				- DataCite:Distributor
+				- DataCite:Editor
+				- DataCite:HostingInstitution
+				- DataCite:Producer
+				- DataCite:ProjectLeader
+				- DataCite:ProjectManager
+				- DataCite:ProjectMember
+				- DataCite:RegistrationAgency
+				- DataCite:RegistrationAuthority
+				- DataCite:RelatedPerson
+				- DataCite:Researcher
+				- DataCite:ResearchGroup
+				- DataCite:RightsHolder
+				- DataCite:Sponsor
+				- DataCite:Supervisor
+				- DataCite:WorkPackageLeader
+				- DataCite:Other
+				- CRediT:conceptualization
+				- CRediT:data-curation
+				- CRediT:formal-analysis
+				- CRediT:funding-acquisition
+				- CRediT:investigation
+				- CRediT:methodology
+				- CRediT:project-administration
+				- CRediT:resources
+				- CRediT:software
+				- CRediT:supervision
+				- CRediT:validation
+				- CRediT:visualization
+				- CRediT:writing-original-draft
+				- CRediT:writing-review-editing
 	*/
 	typedef structure {
 		string contributor_type;
@@ -880,23 +948,42 @@ module Workspace {
 		- resource_type
 		- titles (one or more required)
 		- contributors (one or more required)
-		- versioning information: if the resource does not have an explicit version number, one or more dates should be supplied, ideally the date of resource publication and the last update (if applicable).
+		- versioning information: if the resource does not have an explicit version number, one or more dates should be supplied: ideally the date of resource publication and the last update (if applicable).
 
 		string identifier - persistent unique identifier for the resource.
-			Should be in the format
-			<database name>:<identifier within database>
-			e.g.	DOI:10.46936/10.25585/60000745
-				CrossRef:IsManifestationOf
-				GO:0005456
-				HGNC:7470
+			Should be in the format <database name>:<identifier within database>
+			Examples:
+				- DOI:10.46936/10.25585/60000745
+				- GO:0005456
+				- HGNC:7470
+
 		string license (optional) - usage license for the resource.
+			Examples:
+				- CC0
+				- Creative Commons 4.0
+				- MIT
+
 		string resource_type - the type of resource being represented.
-		string version (optional) - the version of the resource.
+			Valid 'resource_type' values:
+				- dataset
+				- record
+
+		string version (optional) - the version of the resource. This must be an absolute version, not a relative version like 'latest'.
+			Examples:
+				- 5
+				- 1.2.1
+				- 20220405
+
 		list<Contributor> contributors (optional) - a list of people and/or organizations who contributed to the resource.
+
 		list<EventDate> dates - a list of relevant lifecycle events for the resource.
+
 		list<FundingReference> funding (optional) - zero or more funding sources for the resource.
+
 		list<PermanentID> related_identifiers (optional) - zero or more PIDs related to the resource.
+
 		list<Title> titles - one or more titles for the resource.
+
 	*/
 	typedef structure {
 		string identifier;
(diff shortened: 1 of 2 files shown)


workspace.spec
```diff
@@ -572,6 +572,359 @@ module Workspace {
 		string description;
 	} ProvenanceAction;
 
+	/* Credit Metadata-related objects and fields
```
ialarmedalien, line 575: where's the best place to put this info?

workspace.spec
```diff
@@ -572,6 +572,359 @@ module Workspace {
...
+		CM is stored in a list as part of the object provenance, but unlike other provenance information, which is immutable, CM can be updated.
+
+		Credit metadata is added after object creation, and requires the user to be a workspace admin. Username, timestamp, and credit metadata schema version are stored along with the CM, which allows tracking of CM changes.
+
+		To update existing CM, a new set of credit metadata is added to the existing list.
+
+		When requesting credit metadata, the last (i.e. most recent) entry in the list is returned.
+
+		In the following entries, "resource" is used to refer to the workspace object that the CM pertains to.
+	*/
```
ialarmedalien, line 588: this blurb is not written in a particularly fluid manner -- any suggestions to make it sound a bit less lumpen gladly received

workspace.spec
```diff
@@ -572,6 +572,359 @@ module Workspace {
...
+
+	/* EventDate
+
+		Represents an event in the lifecycle of a resource and the date it occurred on.
+		These objects are used to capture information about
```
ialarmedalien, line 593: oops.

is another sentence necessary or will the first sentence suffice?

workspace.spec
```diff
@@ -572,6 +572,359 @@ module Workspace {
...
+			"translated_title"
+			"other"
+	*/
+	typedef structure {
+		string title_string;
+		string title_language;
+		string title_type;
+	} Title;
+
+	/* Contributor
```
ialarmedalien, line 773: not yet added under `workspace.database.provenance`

workspace.spec
```diff
@@ -572,6 +572,359 @@ module Workspace {
...
+		string name;
+		string given_name;
+		string family_name;
+		string contributor_id;
+		list<Organization> affiliations;
+		list<string> contributor_roles;
+	} Contributor;
+
+	/*
+		Represents the credit metadata associated with a workspace object.
```
ialarmedalien, line 858: not yet added

workspace.spec
```diff
@@ -572,6 +572,359 @@ module Workspace {
...
+		string schema_version - version of the credit metadata schema used
+		CreditMetadata credit_metadata - the credit metadata itself
+	*/
+	typedef structure {
+		username saved_by;
+		epoch timestamp;
+		CreditMetadata credit_metadata;
+		string schema_version;
+	} CreditMetadataContainer;
+
```
ialarmedalien, line 927: N.b. I haven't yet added the functions that will deal with these objects.

workspace.spec
elishawc: updates - "new set" is a little confusing. Implies the prev record is archived and a new record is used. But maybe that is correct, given the line after. So if someone added an author, we wouldn't just append the record w/ a change log, we would archive the prev list, and provide a new copy of the list with the addition.

workspace.spec
```diff
@@ -572,6 +572,359 @@ module Workspace {
...
+		string award_id;
+		string award_title;
+		string award_url;
+	} FundingReference;
+
+	/* Organization
+
+		Represents an organization.
+
+		'organization_name' is required; 'organization_id' is optional.
```
elishawc, line 642: Can we recommend a place to look for these? e.g., ror.org/
Also, if there is somewhere that has "standards" for Org Names, that would useful 🤦

workspace.spec
```diff
@@ -572,6 +572,359 @@ module Workspace {
...
+		The values in the 'relationship_type' field come from controlled vocabularies maintained by Datacite and CrossRef. See the documentation links below for more details.
+
+		DataCite relation types: https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#12b-relationtype
+
+		CrossRef relation types: https://www.crossref.org/documentation/schema-library/markup-guide-metadata-segments/relationships/
+
+		string id - persistent unique ID for an entity. Should be in the format
+			<database name>:<identifier within database>
+			e.g.	DOI:10.46936/10.25585/60000745
+				CrossRef:IsManifestationOf
```
elishawc, line 669: IsManifestationOf... is that right?

workspace.spec
elishawc: This looks good. At the minimum should cover the fields to display the recommended ESIP citation, in case we want to add that to documentation.
https://esip.figshare.com/articles/online_resource/Data_Citation_Guidelines_for_Earth_Science_Data_Version_2/8441816

workspace.spec
ialarmedalien: ror.org is probably as good a place as any to look for organization names -- I'll add that to the documentation

workspace.spec
ialarmedalien: > Implies the prev record is archived and a new record is used. But maybe that is correct, given the line after.

That is correct -- it's just that we're saving all the records in a list, with the last entry in the list being the most recent. To update credit metadata, you would upload a new set of credit metadata, which would get added to the end of the list. The last list item is returned when there is a request for credit metadata. If there was a need to track down when some change was made to the metadata, we can look through the list entries to see when it first appeared.

MrCreosote:
First comment - can the base be changed to a branch specifically for credit provenance development? The other PRs were fine going to develop as the code added was not visible to the API, but if we merge this and then want to release the workspace for some other feature before development is complete the documentation will contain references to incomplete features

workspace.spec
MrCreosote: IIRC we said we were going to expose it in https://ci.kbase.us/services/ws/docs/Workspace.html#typedefWorkspace.ObjectData 
Assuming I'm right, this seems like a good a place as any. If I'm wrong, and it's going in `ProvenanceAction`, then the de facto standard is to put the data structure definition prior to the use

workspace.spec
elishawc: does that require some to have the full credit metadata record to resubmit?

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
 		string description;
 	} ProvenanceAction;
 
+	/* Credit Metadata-related objects and fields
```
MrCreosote, line 575: I'm pretty sure this comment will get dropped by the parser (you could test if it winds up in the html documentation to be sure). If so, you'll want to integrate into a comment for a specific data structure or the module comment

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+
+		Workspace objects have a list of zero or more sets of credit metadata (CM), or citation information for the object. This information allows data imported into KBase to retain the appropriate details for users wishing to cite the data source, and for those who contributed to its creation to be credited for their work.
```
MrCreosote, line 577: Please line wrap @ 100 chars

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+
+		CM is stored in a list as part of the object provenance, but unlike other provenance information, which is immutable, CM can be updated.
+
+		Credit metadata is added after object creation, and requires the user to be a workspace admin. Username, timestamp, and credit metadata schema version are stored along with the CM, which allows tracking of CM changes.
+
+		To update existing CM, a new set of credit metadata is added to the list.
+
+		When requesting credit metadata, the last (i.e. most recent) entry in the list is returned.
```
MrCreosote, line 585: I was under the impression that we would store the CM as a list, but only the most recent CM would be exposed to the user. In effect, the users don't see a list, they only see the most recent update. The list might be accessible via a specialized admin command or just direct db access for historical purposes. Is this wrong?

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+
+		Recommended resources for organization identifiers include:
+			- Research Organization Registry, http://ror.org
+			- International Standard Name Identifier, https://isni.org
+			- Crossref Funder Registry, https://www.crossref.org/services/funder-registry/
+
+		Some organizations may have a digital object identifier (DOI).
+
+		string funder_id (optional) - persistent unique identifier for the funder in the format
+			<database name>:<identifier within database>
```
MrCreosote, line 632: examples for the funder_id and funder_name fields would be helpful

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+
+	/* EventDate
+
+		Represents an event in the lifecycle of a resource and the date it occurred on.
+
+		See https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#8-date for more information on the events (below).
+
+		Both event and date are required fields.
+
+		string date - the date associated with the event. The date may be in the format
```
MrCreosote, line 598: I've started moving away from recapitulating the type in the comment - seems redundant. Up to you whether you want to do that here. Otherwise, at some point in the future, I'll probably do a clean sweep of the spec

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		Credit metadata is added after object creation, and requires the user to be a workspace admin. Username, timestamp, and credit metadata schema version are stored along with the CM, which allows tracking of CM changes.
+
+		To update existing CM, a new set of credit metadata is added to the list.
+
+		When requesting credit metadata, the last (i.e. most recent) entry in the list is returned.
+
+		In the following entries, "resource" is used to refer to the workspace object that the CM pertains to.
+	*/
+
+	/* EventDate
```
MrCreosote, line 590: I assume this compiles?

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		Represents an organization.
+
+		Recommended resources for organization identifiers and canonical organization names include:
+			- Research Organization Registry, http://ror.org
+			- International Standard Name Identifier, https://isni.org
+			- Crossref Funder Registry, https://www.crossref.org/services/funder-registry/
+
+		For example, the US DOE would be entered as:
+			organization_name: United States Department of Energy
+			organization_id:   ROR:01bj3aw27
```
MrCreosote, line 657: very readable

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		string description;
+		string relationship_type;
+	} PermanentID;
+
+	/* Title
+		Represents the title or name of a resource.
+
+		The 'title_string' field is required; if no value is supplied for 'title_type', it
+		defaults to 'title'. If the 'title_type' is set to 'translated_title', the appropriate
+		BCP-47 tag must be supplied in the 'title_language' field.
```
MrCreosote, line 773: I don't think this constraint is coded in the class is it?

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+
+		string title_string - the resource title
+		string title_language (optional) - language that the title is in, as a IETF BCP-47 tag.
+			Examples: 'fr'; 'jp-JP'; 'zh-Hant-CN'; 'en-Latn-GB'; 'mn-Cyrl'.
```
MrCreosote, line 777: Should note that the workspace will check the syntax but not the semantics

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		Contributors must have a 'contributor_type', either 'person' or 'organization'.
+
+		There must also be an identifier or name of some sort for the contributor.
+
+		If the contributor is an organization, either the 'contributor_id' or
+		the 'name' field must be populated.
+
+		An individual can be represented by either 'given_name' and 'family_name'; a 'name'
+		(if the name is not in the binomial format); or a `contributor_id`, such as an
+		ORCID.
```
MrCreosote, line 804: I'm wondering
* why not just use the name field and drop *_name
* if name should be required regardless of whether an ID is available. That'd be better for display purposes anyway, I'd think

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+
+		The 'contributor_role' field takes values from the Datacite and CRediT contributor
+		roles vocabularies. For more information on these resources and choosing the
+		appropriate roles, please see the following links:
+
+		Datacite contributor roles: https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#7a-contributortype
+
+		CRediT contributor role taxonomy: https://credit.niso.org
```
MrCreosote, line 812: It'd be nice if all these orgs with overlapping categories could work together

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		string contributor_id (optional) - persistent unique identifier for the contributor;
+			e.g. an ORCID for an individual, or a ROR ID for an organization.
+		string given_name (optional) - first name, for individuals whose names can be split
+			into given and surnames
+		string family_name (optional) - family name, for individuals whose names can be
+			split into given and surnames.
+		string name (optional) - contributor name, in cases where the contributor is an
+			organization or a person whose name cannot easily fit into the given/family
+			name structure.
+		list<Organization> affiliations (optional) - list of organizations with which the
```
MrCreosote, line 824: Should this be empty if the contributor is an org?

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		string version;
+		string resource_type;
+		list<Contributor> contributors;
+		list<EventDate> dates;
+		list<FundingReference> funding;
+		list<PermanentID> related_identifiers;
+		list<Title> titles;
+	} CreditMetadata;
+
+	/* CreditMetadataContainer
```
MrCreosote, line 913: Given that at least the first version of this feature is going to be admin edits only, and the admin interface just takes `UnspecifiedObject`s, I don't think there's a need for this structure

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		username saved_by - the user who added this entry
+		epoch timestamp - timestamp for the addition of this credit metadata
+		string schema_version - version of the credit metadata schema used
+		CreditMetadata credit_metadata - the credit metadata itself
+	*/
+	typedef structure {
+		username saved_by;
+		epoch timestamp;
+		CreditMetadata credit_metadata;
+		string schema_version;
```
MrCreosote, line 944: I'm having qualms about having a schema version. Any backward incompatible change (or even forwards compatible change) should result in at workspace version bump (accessible programatically and cheaply via `ver()`). Is that not enough?

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		string contributor_id;
+		list<Organization> affiliations;
+		list<string> contributor_roles;
+	} Contributor;
+
+	/*
+		Represents the credit metadata associated with a workspace object.
+
+		Required fields are:
+		- identifier
```
MrCreosote, line 879: `id` or `*_id` is used for semantically equivalent fields in other areas, do you want to be consistent?

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		- resource_type
+		- titles (one or more required)
+		- contributors (one or more required)
+		- versioning information: if the resource does not have an explicit version number, one or more dates should be supplied, ideally the date of resource publication and the last update (if applicable).
+
+		string identifier - persistent unique identifier for the resource.
```
MrCreosote, line 885: `the resource` = the source data for the workspace object? If so I'd make that explicit

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+			Should be in the format
+			<database name>:<identifier within database>
+			e.g.	DOI:10.46936/10.25585/60000745
+				CrossRef:IsManifestationOf
+				GO:0005456
+				HGNC:7470
+		string license (optional) - usage license for the resource.
+		string resource_type - the type of resource being represented.
```
MrCreosote, line 893: Definitely need more explanation / examples here, I'd have no idea what was required here

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+		- resource_type
+		- titles (one or more required)
+		- contributors (one or more required)
+		- versioning information: if the resource does not have an explicit version number, one or more dates should be supplied, ideally the date of resource publication and the last update (if applicable).
+
+		string identifier - persistent unique identifier for the resource.
+			Should be in the format
+			<database name>:<identifier within database>
+			e.g.	DOI:10.46936/10.25585/60000745
+				CrossRef:IsManifestationOf
```
MrCreosote, line 889: Is this actually a resource ID that could occur...?!

workspace.spec
```diff
@@ -572,6 +572,378 @@ module Workspace {
...
+				GO:0005456
+				HGNC:7470
+		string license (optional) - usage license for the resource.
```
MrCreosote, line 892: Is this the full license text or just the name of the license? For the latter, is there a CV or something like that?

workspace.spec
MrCreosote: Note that if we don't include it as a part of the provenance data structure, we'll need to do a separate size check (and have a separate, hopefully very small, size limit)

workspace.spec
ialarmedalien: No, that is correct -- I just haven't worded my description of it very well. I think it may be better to hide the implementation details from the spec file and just reflect the public API here, since both you and Elisha have had questions after reading that description!

MrCreosote: That makes sense to me. Any docs for admin functions that can access older entries would go here: https://ci.kbase.us/services/ws/docs/administrationinterface.html

workspace.spec
ialarmedalien: Remedied: https://github.com/kbase/workspace_deluxe/pull/659

workspace.spec
```diff
@@ -572,6 +572,468 @@ module Workspace {
...
+
+		Some organizations may have a digital object identifier (DOI).
+
+		string funder_id (optional) - persistent unique identifier for the funder in the format
+			<database name>:<identifier within database>
+			Examples:
+				- ROR:04xm1d337
+				- ISNI:0000000405337147
+
+		string funder_name - common name for the funder
```
ialarmedalien, line 643: not sure what this strange line wrapping is about!

workspace.spec
ialarmedalien: I've gone through and added examples for as many fields as possible

workspace.spec
MrCreosote: :+1:

workspace.spec
ialarmedalien: We discussed this quite a bit in the CE group.

> why not just use the name field and drop *_name

majority preferred to keep the given/family name structure (despite problems inherent in assuming person names conform to this syntax, etc.).

> if name should be required regardless of whether an ID is available.

👍 

> That'd be better for display purposes anyway, I'd think

Ideally at some point -- either in the credit engine or when displaying the contributors -- this data could be dynamically sourced from providers like ORCID.

workspace.spec
ialarmedalien: I'd still want to have the version of the credit metadata schema easily available, though -- otherwise you would need to have a mapping of workspace version to credit metadata schema version.

workspace.spec
ialarmedalien: Once text changes are finalised I will implement line wrap throughout.

workspace.spec
ialarmedalien: Clarified

workspace.spec
ialarmedalien: For adding/updating records, though, since the workspace API + data structures are going to be so closely linked to the CM schema, the CM schema version can be added automatically.

workspace.spec
ialarmedalien: I'm going to create a new controlled vocabulary that unites these two efforts and is the new standard CV for contributor roles.

workspace.spec
ialarmedalien: just to check: if it goes under ObjectData, will there be a single version for all copies and copies of copies (of copies of copies, etc.) that get made of the object?

workspace.spec
ialarmedalien: @elishawc what do you think on this issue? For at least one data source (JGI), there's a custom license that basically says that you should cite the data source if you're going to use it. Even though that is probably covered by an existing license (e.g. one of the Creative Commons licenses), it's not expressed as such on the JGI website so we should stay faithful to the source.

workspace.spec
ialarmedalien: yep. I've updated the explanation to clarify.

workspace.spec
ialarmedalien: Removed

ialarmedalien:
Re: comment about merging -- I can create a feature branch (or use this as the feature branch) and delay the merging of the spec and related files until the workspace API matches what is in the spec.

workspace.spec
ialarmedalien: Removed.

workspace.spec
MrCreosote: You could implement it either way, depending on what you wanted. The database schema and API schemas don't really have much to do with each other

MrCreosote: And FWIW we could keep it in the internal Provenance data structure internally and use the existing size check

MrCreosote:
I would make a specific feature branch so we can merge this into it when we're happy with it and we don't clutter future reviews with it

workspace.spec
MrCreosote: :+1:

workspace.spec
MrCreosote: I'm saying I don't think you need a schema version at all. When you say "the credit schema version is changing" what you're really saying is "the workspace API is changing". We don't have a schema version for provenance, say, and that's been fine for 9 years.

The way we manage that kind of thing would be
a) avoid backwards incompatible changes like the plague, and if that's impossible
b) have a toggle so the user can choose which version of the API data structures they want, with the understanding that older (default) versions may have less information.

workspace.spec
MrCreosote: Are you serious or kidding? I think you're kidding but if you are serious I wouldn't do that

workspace.spec
MrCreosote: :+1:

workspace.spec
MrCreosote: :+1: for requiring the name

I'm still iffy about the two different name fields. Couldn't you split `name` on white space and if there's two parts assume it's family and given?

workspace.spec
MrCreosote: :+1:

workspace.spec
MrCreosote: :+1:

workspace.spec
MrCreosote: Could you define what expected for a `dataset` vs a `record`

workspace.spec
MrCreosote: :+1:

workspace.spec
ialarmedalien: I'll remove them all in the next commit.

MrCreosote: If it's just the types for your additions that great, but if you're talking about the entire spec could we please do that in a separate PR?

workspace.spec
MrCreosote: Maybe a URL to a license?

elishawc: super valid Q, and I should probably check in with Grace and KJ on plans for that, but we could just link to their policy page for now. That was what Grace asked when we updated our T&Cs. 
https://jgi.doe.gov/user-programs/pmo-overview/policies/

workspace.spec
ialarmedalien: Done for the entries in this PR.

workspace.spec
elishawc: we could reference the datacite documentation, which contains a variety of resource types. https://support.datacite.org/docs/what-are-the-resource-types-for-datacite-dois (too bad they don't define them there). @ialarmedalien - why did you pick dataset and record?

workspace.spec
ialarmedalien: FWIW I agree. I'll bring it up at the CE meeting tomorrow and see if minds can be changed.

workspace.spec
ialarmedalien: After further discussion with Elisha, this is going to be reverted to 'dataset', since that is the only use case we will be dealing with for the foreseeable future.

workspace.spec
ialarmedalien: Since this is the highest of the high mucky mucks of identifiers related to this resource, I'm giving it the honour of writing its name out in full.

workspace.spec
ialarmedalien: For now, this field can be a string or an URL. If anything changes prior to implementing the class, the spec can be updated accordingly.

workspace.spec
MrCreosote: :+1:

workspace.spec
MrCreosote: :+1:

workspace.spec
MrCreosote: So if it looks like a url it'll be interpreted as a url IOW?

workspace.spec
MrCreosote: in that case I have to ask if the field is necessary at all

elishawc: I would vote yes, as this might need to expand to other controlled vocabulary as the platform matures. e.g., KBase Narratives as preprints

workspace.spec
```diff
@@ -572,6 +572,466 @@ module Workspace {
...
+
+	/* CreditMetadataEntry
+
+		Container for an instance of credit metadata; returned by the workspace when credit metadata is requested.
+
+		All fields will be populated.
+
+		credit_metadata - the credit metadata itself
+
+		credit_metadata_schema_version - version of the credit metadata schema used
```
MrCreosote, line 1020: Why is the schema version in `CreditMetadata` and `CreditMetadataEntry`?

workspace.spec
MrCreosote: I  unresolved this as it looks like there's still references to the list, e.g. `Workspace objects have a list of zero or more sets of credit metadata (CM), or citation information for the object.`

workspace.spec
MrCreosote: It looks like it's integrated into the EventDate comment now, which I'm not sure makes sense - I'd suggest the root data structure in the tree

workspace.spec
MrCreosote: :+1:

workspace.spec
MrCreosote: Can it be added later when it's needed, or can it default to dataset so everyone who uses the api doesn't have to fill in a field where there's no other choice

elishawc: I defer to @ialarmedalien and the work to version, but def can be set to default = dataset

workspace.spec
ialarmedalien: OK, will move it.

workspace.spec
elishawc: I def think we want to have transparent versioning of the credit metadata. This isn't really linked to workspace API updates, this is more linked to external changes in the publishing infrastructure, changes in the DataCite schema, etc. This is implemented in our workspace, but isn't really about us :)

workspace.spec
ialarmedalien: I want `credit_metadata_schema_version` to be submitted by the user when they add the credit metadata, i.e.

```json
{
    "credit_metadata_schema_version": "1.1.0",
    "credit_metadata": { ... }
}
```

When a submission is made, `saved_by` and `timestamp` fields will be added automatically, and the saved data structure will look like this:

```json
{
    "credit_metadata_schema_version": "1.1.0",
    "saved_by": "user_name_here",
    "timestamp": 1234567890,
    "credit_metadata": { ... }
}
```

This structure is what should be returned when users request credit metadata for a WS object.

I think I misinterpreted something you said and that I need to reinstate the spec for this data structure:
```json
{
    "credit_metadata_schema_version": "1.1.0",
    "credit_metadata": { ... }
}
```
and undo the addition of the `credit_metadata_schema_version` from `CreditMetadata`.

workspace.spec
ialarmedalien: Yes, it can definitely be set to default to dataset. I would definitely want to keep the field in because it's an important part of the schema.

workspace.spec
MrCreosote: Looks like the schema version was removed from `CreditMetadata` so the OQ is moot.

> What would be your proposal for a function for adding credit metadata to an object, and where would it go -- in this workspace.spec file or elsewhere?

I would use the extant `administer()` method.

workspace.spec
MrCreosote: >		CM is stored as part of the object provenance, but unlike other provenance information, which is immutable, CM can be updated.
>		Credit metadata is added after object creation, and requires the user to be a workspace admin. Username, timestamp, and credit metadata schema version are stored along with the CM, which allows tracking of CM changes.
>		The API does not support updates to select fields; the existing CM for a resource should be fetched, the relevant fields updated, and the resulting structure sent as the input to update credit metadata.

Maybe add this later when it becomes true? I think for the first version it's fixed yeah?

.classpath
```diff
@@ -8,103 +8,103 @@
...
 		</attributes>
 	</classpathentry>
-	<classpathentry kind="lib" path="/jars/lib/jars/apache_commons/commons-logging-1.1.1.jar"/>
-	<classpathentry kind="lib" path="/jars/lib/jars/ini4j/ini4j-0.5.2.jar"/>
-	<classpathentry kind="lib" path="/jars/lib/jars/jetty/jetty-all-7.0.0.jar"/>
-	<classpathentry kind="lib" path="/jars/lib/jars/jna/jna-3.4.0.jar"/>
-	<classpathentry kind="lib" path="/jars/lib/jars/servlet/servlet-api-2.5.jar"/>
-	<classpathentry kind="lib" path="/jars/lib/jars/syslog4j/syslog4j-0.9.46.jar"/>
-	<classpathentry kind="lib" path="/jars/lib/jars/apache_commons/commons-collections-3.2.1.jar" sourcepath="/jars/lib/jars/apache_commons/commons-collections-3.2.1-sources.jar">
+	<classpathentry kind="lib" path="/Users/gwg/code/kbase/jars/lib/jars/apache_commons/commons-logging-1.1.1.jar"/>
```
MrCreosote, line 11: This breaks the classpath file for everyone

lib/biokbase/workspace/baseclient.py
```diff
@@ -125,7 +121,7 @@ def __init__(
             self, url=None, timeout=30 * 60, user_id=None,
             password=None, token=None, ignore_authrc=False,
             trust_all_ssl_certificates=False,
-            auth_svc='https://kbase.us/services/auth/api/legacy/KBase/Sessions/Login',
+            auth_svc='https://kbase.us/services/authorization/Sessions/Login',
```
MrCreosote, line 124: It looks like this was compiled with a very old version of kb-sdk, this is a reversion

MrCreosote (changes requested):

.classpath
ialarmedalien: Accidental commit, now removed

lib/biokbase/workspace/baseclient.py
ialarmedalien: Ditto.

workspace.spec
ialarmedalien: Elisha is firm on wanting to have given / family names.

.classpath
MrCreosote: :+1:

lib/biokbase/workspace/baseclient.py
MrCreosote: :+1:

workspace.spec
ialarmedalien: Resolved by in-person meeting.

workspace.spec
ialarmedalien: Removed that reference.

workspace.spec
```diff
@@ -837,25 +836,21 @@ module Workspace {
...
+		a 'name'.
 
-		If the contributor is an organization, the 'name' field must be populated.
-
-		A person must have either 'given_name' and 'family_name' or a 'name'
-		if the name cannot be represented in the given/family name format. Citations will
-		be generated using the initials from the given name(s) and the family name(s).
+		The 'credit_name' field is used to store the name of a person as it would appear in
+		a citation. If there is no 'credit_name' supplied, the 'name' field would be used
+		in citations.
```
ialarmedalien, line 844: New -- use `credit_name` to represent a person's name as it would appear in a citation.

workspace.spec
ialarmedalien: Now using name / credit_name instead

workspace.spec
ialarmedalien: Removed until it becomes true

MrCreosote: :+1:

workspace.spec
MrCreosote: :+1:

workspace.spec
MrCreosote: specifically, the credit schema will be checked into the workspace and the classes in the workspace will be kept in sync with that schema (with appropriate documentation so that new developers know this), thus removing a circular dependency (e.g. the credit engine depends on the workspace, which depends on a spec stored in the CE to define its class structure for credit data

MrCreosote: The rationale is that the schemas at 3rd party credit sites change fairly rapidly, and the credit schema here is based on those schemas. This does mean, however, that given multiple versions of the credit schema will be stored in the workspace at the same time, any API and internal classes dealing with credit must be able to represent any version of the schema.

workspace.spec
MrCreosote: Can we note the default in the spec?

workspace.spec
MrCreosote: This has been done, I'll recheck on the final review

workspace.spec
MrCreosote: > If there is no 'credit_name' supplied, the 'name' field would be used in citations.

Hmm, this seems like it's virtually certain to be wrong. Are you sure you don't want to require both?

workspace.spec
MrCreosote: :+1:

workspace.spec
ialarmedalien: Done.

workspace.spec
ialarmedalien: If it's not clear from the `name` field what the most appropriate way to format it for citation is, making the `credit_name` field mandatory would mean the possibility of the `credit_name` being wrong because who/whatever was creating the entry would need to put something in that field. If the field is optional, the name might appear in an inappropriate format in the citation, but the `credit_name` field could be updated without needing to investigate why it had been entered in an incorrect format in the first place.

workspace.spec
ialarmedalien: I've left the data structure where it is as the plan is to expose it as part of the workspace object data.

workspace.spec
MrCreosote: fair enough

workspace.spec
```diff
@@ -572,6 +572,497 @@ module Workspace {
...
+		list<EventDate> dates;
+		list<FundingReference> funding;
+		list<PermanentID> related_identifiers;
+		list<Title> titles;
+	} CreditMetadata;
+
+	/* CreditMetadataEntry
+
+		Container for the credit metadata for a workspace object, returned by the workspace
+		when credit metadata is requested.
```
MrCreosote, line 1033: This is a bit off but not really important - when you get the object data you get CM whether you ask for it or not.

workspace.spec
```diff
@@ -572,6 +572,497 @@ module Workspace {
...
+		string identifier;
+		string license;
+		string resource_type;
+		string version;
+		list<Contributor> contributors;
+		list<EventDate> dates;
+		list<FundingReference> funding;
+		list<PermanentID> related_identifiers;
+		list<Title> titles;
+	} CreditMetadata;
```
MrCreosote, line 1028: We should decide on and document a maximum size for the CM when serialized

MrCreosote (commented):
Looking good, just a few things. I'm assume this compiles without errors

Note to self - the CM type doesn't exist yet, but when it does, there's some field default expectations that should be checked for in the PR review

workspace.spec
```diff
@@ -572,6 +572,497 @@ module Workspace {
...
+		string id;
+		string description;
+		string relationship_type;
+	} PermanentID;
+
+	/* Title
+		Represents the title or name of a resource.
+
+		The 'title_string' field is required; if no value is supplied for 'title_type', it
+		defaults to 'title'. If the 'title_type' is set to 'translated_title', the
```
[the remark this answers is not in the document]
ialarmedalien: the name was from DataCite's schema -- I agree it isn't all that great, but at least the definition makes it clear what it is

workspace.spec
ialarmedalien: Yeah, I will edit it to say that it's returned when object data is requested.

ialarmedalien: On second thoughts, I'm just going to delete the second half of that sentence.

ialarmedalien:
@MrCreosote looks like there's still a change requested somewhere amongst the multitude of comments. Can you dismiss it and approve the PR (assuming it is now in a semi-reasonable state)?

workspace.spec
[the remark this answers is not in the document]
MrCreosote: Maybe add a comment explicitly explaining the above? I'm sure I won't be the first person to be confused

workspace.spec
MrCreosote: I was thinking we should do that in this PR - did you want to delay?

ialarmedalien: yes -- was going to put in the documentation when the size check is implemented

workspace.spec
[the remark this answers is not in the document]
ialarmedalien: Comment added!

workspace.spec
MrCreosote: Ok, as long as it winds up in the spec at some point

:+1:

workspace.spec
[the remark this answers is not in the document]
MrCreosote: :+1:

MrCreosote (approved):

PR #659, opened 2023-01-06
CE-151: add title language constraint

ialarmedalien:
ensure titleLanguage is specified if titleType is translated_title


MrCreosote (approved):
LGTM

PR #663, opened 2023-01-30
Migrate to Java 11

ialarmedalien:
Update Dockerfile and relevant build files to compile and run on Java 11


.github/workflows/test.yml
```diff
@@ -21,14 +21,15 @@ jobs:
...
         include:
-          - java: '8'
+          - java: '11'
             mongo: 'mongodb-linux-x86_64-3.6.23'
             minio: '2019-05-23T00-29-34Z'
-            wired_tiger: 'false'
+            wired_tiger: 'true'
             ant_test: 'test_quick_coverage'
-          - java: '11'
+          - java: '17'
```
ialarmedalien, line 32: these tests fail - this section could be considered aspirational

Dockerfile
```diff
@@ -1,22 +1,43 @@
-FROM kbase/sdkbase2 as build
+FROM eclipse-temurin:11-jdk as build
```
ialarmedalien, line 1: The official openjdk images have been deprecated, with a list of recommendations for alternatives given. Eclipse seemed as good as any to choose.

Other possible build images:
- amazoncorretto (runs on Amazon linux)
- ibmjava
- ibm-semeru-runtimes
- kbase/sdkbase2 (bleh, would need java 11 installing)

Dockerfile
```diff
@@ -1,22 +1,43 @@
...
-RUN pip install configobj && \
-    cd /tmp && \
-    git clone https://github.com/kbase/jars && \
-    cd workspace_deluxe && \
+RUN cd workspace_deluxe && \
     make docker_deps
 
-FROM kbase/kb_jre
+# updated/slimmed down version of what's in kbase/kb_jre
+FROM ubuntu:18.04
```
ialarmedalien, line 13: moved what was done on the kbase/kb_jre image into this Dockerfile

Dockerfile
```diff
@@ -1,22 +1,43 @@
...
 # These ARGs values are passed in via the docker build command
 ARG BUILD_DATE
 ARG VCS_REF
 ARG BRANCH=develop
 
+ENV DOCKERIZE_VERSION linux-amd64-v0.6.1
+ENV TOMCAT_VERSION tomcat8
+USER root
+
+RUN mkdir -p /var/lib/apt/lists/partial && \
```
ialarmedalien, line 24: this was done on the kbase/kb_jre image

deployment/conf/.templates/setenv.sh.templ
```diff
@@ -1,17 +1,15 @@
 #!/bin/sh
-#
-
 CATALINA_HOME=/usr/share/tomcat8
 
 # The first existing directory is used for JAVA_HOME (if JAVA_HOME is not
 # defined in $DEFAULT)
-JDK_DIRS="/usr/lib/jvm/java-8-openjdk-amd64 /usr/lib/jvm/jre-8-oracle-x64"
+JDK_DIRS="/usr/lib/jvm/java-11-openjdk-amd64 /usr/lib/jvm/java-8-openjdk-amd64"
```
ialarmedalien, line 6: updating for Java 11

.github/workflows/test.yml
```diff
@@ -21,14 +21,15 @@ jobs:
...
     strategy:
+      fail-fast: false
       matrix:
         include:
-          - java: '8'
+          - java: '11'
             mongo: 'mongodb-linux-x86_64-3.6.23'
             minio: '2019-05-23T00-29-34Z'
-            wired_tiger: 'false'
+            wired_tiger: 'true'
```
MrCreosote, line 30: One of the tests needs to run with wired tiger off since it's off in prod

.github/workflows/test.yml
```diff
@@ -21,14 +21,15 @@ jobs:
   workspace_deluxe_tests:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
MrCreosote, line 24: Most important question - have you fired up the server in a docker container and made sure it works? I think minimum config would be a local mongo and pointing it at CI auth

.gitignore
```diff
@@ -0,0 +1,8 @@
+/bin/
```
MrCreosote, line 1: Can you please not check in `.gitignore`? They become masses of cruft with everyone checking in their own little personal ignore sets and effectively claiming directories and files

I think you may need to force push *gasp* to get rid of it completely

Dockerfile
```diff
@@ -1,22 +1,43 @@
-FROM kbase/sdkbase2 as build
+FROM eclipse-temurin:11-jdk as build
+
+WORKDIR /tmp
+RUN apt-get update -y && \
+        apt-get install -y tzdata && apt-get install -y ca-certificates python3-sphinx ant git && \
```
MrCreosote, line 5: Why 2 different apt-get invocations on the same line?

.github/workflows/test.yml
```diff
@@ -21,14 +21,15 @@ jobs:
   workspace_deluxe_tests:
     runs-on: ubuntu-latest
     strategy:
```
MrCreosote, line 23: There are probably documentation changes that are needed to note that that workspace now requires java 11. In the readme for sure, but search the docs folder to see if there's anywhere else>

build.xml
```diff
@@ -201,8 +201,8 @@
     <javac
         destdir="${classes}"
         includeantruntime="false"
-        target="1.8"
-        source="1.8"
+        target="11"
+        source="11"
```
MrCreosote, line 205: Is this the only place where target and source versions are specified?

MrCreosote (changes requested):

.github/workflows/test.yml
MrCreosote: Removed, I don't want the tests to fail every time we PR. Maybe just have another java 11 test with wired tiger off and mongo version 3.6.13 which is in prod. The test above can be "cutting edge" mongo weeps histrionically

Dockerfile
MrCreosote: We're running GHA tests with temurin so this is a good choice

Dockerfile
MrCreosote: Everything below is a straight C&P?

deployment/conf/.templates/setenv.sh.templ
MrCreosote: Shouldn't the 2nd dir also be java 11?

MrCreosote: Shouldn't the 2nd dir be `/usr/lib/jvm/jre-11-openjdk-x64`? Let's just ditch 8 completely if we're moving to 11

Dockerfile
```diff
@@ -1,22 +1,43 @@
...
+FROM ubuntu:18.04
 
 # These ARGs values are passed in via the docker build command
 ARG BUILD_DATE
 ARG VCS_REF
 ARG BRANCH=develop
 
+ENV DOCKERIZE_VERSION linux-amd64-v0.6.1
+ENV TOMCAT_VERSION tomcat8
+USER root
```
MrCreosote, line 22: Well this makes me nervous

Dockerfile
ialarmedalien: Here's the [kb_jre image Dockerfile](https://github.com/kbase/kb_jre/blob/master/Dockerfile).

The only changes that I made were to turn the tomcat and dockerize versions into env vars for ease of updating.

.github/workflows/test.yml
ialarmedalien: Can we come up with a minimal test (or test set) that could be run on the built image to ensure that it is functional? That would be better, so that there's some sort of framework in place for the next migration (in twenty years' time).

Dockerfile
ialarmedalien: It's copied from the kbase/kb_jre Dockerfile. Presumably the root user is required to set up all the highly important directories and permissions that follow.

deployment/conf/.templates/setenv.sh.templ
ialarmedalien: I was in several minds about what to do with this script (template). The existing script is a modification of a `setenv.sh` script installed as part of tomcat, and the current version changes the tomcat script by altering the JDK directories to look for JDK 8 and adding in run time parameters for tomcat.

original tomcat script:
```sh
#!/bin/sh
#

CATALINA_HOME=/usr/share/tomcat8

# The first existing directory is used for JAVA_HOME (if JAVA_HOME is not
# defined in $DEFAULT)
JDK_DIRS="/usr/lib/jvm/java-7-openjdk-amd64 /usr/lib/jvm/jre-7-oracle-x64"

# Look for the right JVM to use
for jdir in $JDK_DIRS; do
    if [ -r "$jdir/bin/java" -a -z "${JAVA_HOME}" ]; then
        JAVA_HOME_TMP="$jdir"
        # checks for a real JDK like environment, needed to check if 
        # really the java-gcj-compat-dev package is installed
        if [ -r "$jdir/bin/jdb" ]; then
            JAVA_HOME="$JAVA_HOME_TMP"
        fi
    fi
done

# Default Java options
if [ -z "$JAVA_OPTS" ]; then
        JAVA_OPTS="-Djava.awt.headless=true"
fi
```
(n.b. `JAVA_HOME` is already set to `/usr/local/openjdk-8`; neither of the paths in `JDK_DIRS` exist)

version that appears on the current workspace image:
```sh
#!/bin/sh
#

# CATALINA_HOME=/usr/share/tomcat8

# The first existing directory is used for JAVA_HOME (if JAVA_HOME is not
# defined in $DEFAULT)
JDK_DIRS="/usr/lib/jvm/java-8-openjdk-amd64 /usr/lib/jvm/jre-8-oracle-x64"

# Look for the right JVM to use
for jdir in $JDK_DIRS; do
    if [ -r "$jdir/bin/java" -a -z "${JAVA_HOME}" ]; then
        JAVA_HOME_TMP="$jdir"
        # checks for a real JDK like environment, needed to check if
        # really the java-gcj-compat-dev package is installed
        if [ -r "$jdir/bin/jdb" ]; then
            JAVA_HOME="$JAVA_HOME_TMP"
        fi
    fi
done

# Default Java options
if [ -z "$JAVA_OPTS" ]; then
        JAVA_OPTS="-Djava.awt.headless=true -server \
               -Xms1000m -Xmx3000m \
               -XX:+UseG1GC"
fi
```
(note the extra `JAVA_OPTS`)

Even if `JAVA_HOME` was not already set, it wouldn't be changed by this script as there's no `jdb` in the bin directory. I don't know why this content was kept in the script, other than perhaps inertia.

The `setenv.sh` script thus could/should be reduced to these lines:

```sh
# Default Java options
if [ -z "$JAVA_OPTS" ]; then
        JAVA_OPTS="-Djava.awt.headless=true -server \
               -Xms1000m -Xmx3000m \
               -XX:+UseG1GC"
fi
```

build.xml
ialarmedalien: I had a couple of questions about `build.xml` involving that:

1. is there some reason why this is set to `1.6`?
```xml
  <target name="compile_client" description="compile only the client">
    <mkdir dir="${client_classes}"/>
    <javac destdir="${client_classes}" includeantruntime="false" target="1.6" source="1.6"
      debug="true" classpathref="compile.classpath">
```

2. is there a reason why these extra entries are included in the java docs?
```xml
    <javadoc access="protected" author="false" classpathref="compile.classpath"
      destdir="${doc}" nodeprecated="false" nodeprecatedlist="false"
      noindex="false" nonavbar="false" notree="false"
      source="1.8" splitindex="true" use="true" version="true">
      <link href="http://download.oracle.com/javase/8/docs/api/"/>
      <link href="http://fasterxml.github.io/jackson-core/javadoc/2.2.0/"/>
      <link href="http://fasterxml.github.io/jackson-databind/javadoc/2.2.0/"/>
      <fileset refid="client-jdoc-files"/>
      <fileset refid="sources-jdoc-files"/>
    </javadoc>
```
The Java SE link seems reasonable but the jackson-core and jackson-databind links are broken, which makes me wonder if they are being used.

.github/workflows/test.yml
MrCreosote: yeah, it's been in my mental to do list for a long time. I was thinking along the lines of a script that runs docker compose to start mongo, auth in test mode (if that's possible with the current auth docker image, not sure), and the ws, creates a user & token in auth, and then creates and reads a workspace to make sure the service is working.

.gitignore
MrCreosote: yeah, it's idiosyncratic, but IMO it's much better for everyone to maintain their own local ignore files

build.xml
MrCreosote: 1. We can probably delete the compile client target altogether. I put that in there years ago and I don't really think it's needed. It was originally there because ppl were still using java 6 when we upgraded to 7

2. The are some parts of the public api that expose jackson classes and so they're in the docs to link out. Looks like the new docs are at https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/2.2.0/index.html

Dockerfile
MrCreosote: Sorry, does the server run as root or not?

.github/workflows/test.yml
MrCreosote: Later some of the other services could be added (blobstore, handle service, sample service), but I think the tests cover the interactions there pretty well

build.xml
ialarmedalien: Is there anything else in the build.xml file that is no longer needed?

Dockerfile
ialarmedalien: yes, it does

.github/workflows/test.yml
ialarmedalien: perhaps also fetching a page from the doc server?

MrCreosote: Sure, although the only thing that could go wrong there is a messed up web.xml file. The tests show the internals work and the hypothetical docker test would show the tomcat stuff works, so really the only remaining part of the picture is mounting the jar. Wouldn't hurt though

MrCreosote: All that being said, currently there just needs to be, at minimum, some manual test of the server for this PR

build.xml
MrCreosote: Probably, but the whole thing will be deleted once I get gradle working, so probably not worth spending effort on

Dockerfile
MrCreosote: yuck, but not surprised

.github/workflows/test.yml
MrCreosote: :+1:

.github/workflows/test.yml
MrCreosote: :+1:

.gitignore
MrCreosote: Hmm, I saw you force pushed, but the gitignore is still there

build.xml
MrCreosote: :+1:

Dockerfile
MrCreosote: :+1:

.gitignore
MrCreosote: IIUC it needs to be completely blasted from every commit in the PR to not cause issues in the future

build.xml
MrCreosote: For the client, there's a makefile target that calls the ant target that'll need to be deleted, and I think the `client_classes` property in build.xml can be deleted

deployment/conf/.templates/setenv.sh.templ
MrCreosote: If I understand this correctly: If JAVA_HOME and JAVA_OPTS are non-empty strings, this script does nothing (I hate shell script syntax...)?

If so, IMO we should not monkey patch scripts installed by apt-get and instead just set JAVA_HOME and JAVA_OPTS in the docker file, maybe by running a script that sets them based on environment variables as the current script does.

That being said, that might be too much work for this PR. At some point I want to go through the entire dockerfile to update tomcat and simplify the installation and startup as much as possible, but maybe not here

MrCreosote: That being said, we still shouldn't have the references to java 8 in this script, whatever we do

Dockerfile
MrCreosote: Looks like this line is missing?

```
update-ca-certificates && \
```

MrCreosote: Other than that yep, confirmed the same

.gitignore
ialarmedalien: Force push was rebasing on develop. I actually forgot to remove the gitignore from the subsequent commit, which is why it's still around.

Dockerfile
```diff
@@ -1,22 +1,43 @@
-FROM kbase/sdkbase2 as build
+FROM eclipse-temurin:11-jdk as build
+
+WORKDIR /tmp
+RUN apt-get update -y && \
+        apt-get install -y ant git ca-certificates tzdata python3-sphinx && \
```
MrCreosote, line 5: why tzdata btw?

deployment/conf/.templates/deployment.cfg.templ
```diff
@@ -1,9 +1,11 @@
 [Workspace]
 mongodb-database={{ default .Env.mongodb_database "workspace" }}
+mongodb-host={{ default .Env.mongodb_host "ci-mongo" }}
+mongodb-user={{ default .Env.mongodb_user "wsserv" }}
+mongodb-pwd={{ default .Env.mongodb_pwd "" }}
```
ialarmedalien, line 5: these haven't changed, just moved all the mongo config to be in the same place

.github/workflows/test.yml
ialarmedalien: See below for minimal auth2 / workspace integration-type tests.

deployment/conf/.templates/setenv.sh.templ
```diff
@@ -1,28 +1,7 @@
...
-for jdir in $JDK_DIRS; do
-    if [ -r "$jdir/bin/java" -a -z "${JAVA_HOME}" ]; then
-	JAVA_HOME_TMP="$jdir"
-	# checks for a real JDK like environment, needed to check if 
-	# really the java-gcj-compat-dev package is installed
-	if [ -r "$jdir/bin/jdb" ]; then
-	    JAVA_HOME="$JAVA_HOME_TMP"
-	fi
-    fi
-done
```
ialarmedalien, line 20: this does nothing so I've removed it

deployment/conf/.templates/setenv.sh.templ
ialarmedalien: I've removed all but the lines that actually do something, i.e. the lines that set the `JAVA_OPTS`. The tomcat docs suggest customising `setenv.sh` for your environment/needs (see https://tomcat.apache.org/tomcat-8.5-doc/RUNNING.txt) so there's no point in keeping cruft around.

.gitignore
ialarmedalien: the `.gitignore` file should no longer be an issue.

Dockerfile
ialarmedalien: It's gone now, but it was there to prevent issues when installing python - python will prompt you for your timezone, which halts automated builds (e.g. on GHA); installing tzdata first stops that prompting.

.github/workflows/test.yml
ialarmedalien: I think it would be better to split out the docs updates into a separate, followup PR because there are several files involved.

I'd also like to remove a number of files that are no longer used in a followup PR.

.github/workflows/test.yml
```diff
@@ -18,27 +18,51 @@ on:
 # TODO split tests into mongo related & non mongo related. Run the former once.
 jobs:
 
+  workspace_integration_tests:
```
MrCreosote, line 21: Can we call this workspace_container_tests or something like that to disambiguate from the integration tests in the test suite?

Dockerfile
```diff
@@ -1,29 +1,47 @@
...
+WORKDIR /tmp
+RUN apt-get update -y && \
+    apt-get install -y ant git ca-certificates python3-sphinx && \
     git clone https://github.com/kbase/jars && \
     cd workspace_deluxe && \
     make docker_deps
 
-FROM kbase/kb_jre
+# updated/slimmed down version of what's in kbase/kb_jre
+FROM ubuntu:18.04
```
MrCreosote, line 12: Shouldn't this be from some specific jre image for java 11?

Dockerfile
```diff
@@ -1,29 +1,47 @@
...
+ENV TOMCAT_VERSION tomcat8
+USER root
+
+RUN mkdir -p /var/lib/apt/lists/partial && \
+    apt-get update -y && \
+    apt-get install --no-install-recommends -y ca-certificates ${TOMCAT_VERSION}-user libservlet3.1-java wget && \
+    update-ca-certificates && \
+    apt-get clean && \
+    useradd -c "KBase user" -rd /kb/deployment/ -u 998 -s /bin/bash kbase && \
+    mkdir -p /kb/deployment/bin && \
```
MrCreosote, line 33: Looks like these two lines are missing, was that intentional?
```
    mkdir -p /kb/deployment/jettybase/logs/ && \
    touch /kb/deployment/jettybase/logs/request.log && \
```

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
+version: '3.4'
+
+services:
+  workspace:
+    build: .
+    # or use the latest docker image
+    # image: kbase/workspace_deluxe:latest
```
MrCreosote, line 7: Our images are stored on GHA so I don't this would actually work would it?

readme.md
```diff
@@ -37,23 +35,28 @@ isn't available.
 The better but slightly harder alternative is to build the documentation
 yourself.
 
-Building documentation
-----------------------
+## Building documentation
```
MrCreosote, line 38:
```suggestion
### Building documentation
```

scripts/run_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+#
+# script to run the python workspace_container_test.py on GitHub Actions.
+# see .github/workflows/test.yml
+
+docker-compose build
+docker-compose up &
```
MrCreosote, line 7: -d will daemonize the process

scripts/run_tests.sh
```diff
@@ -0,0 +1,12 @@
...
+# ugly way to wait for all the services to be ready
+sleep 60
```
MrCreosote, line 9: In the future could add a loop with the workspace client calling ver() or something

scripts/workspace_container_test.py
```diff
@@ -0,0 +1,82 @@
...
+import requests
+import json
+
+""" workspace_container_test.py
+
+Very simple tests to ensure that local workspace and auth2 servers are functioning correctly.
+Requires the python libraries `pytest` and `requests` to be installed.
+
+To run:
+- move this script to the `workspace_deluxe` (i.e. the parent) directory
```
MrCreosote, line 12: Why require the move rather than run in place?

MrCreosote (commented):
This is fantastic, just some smaller things left

Dockerfile
MrCreosote: :+1:

deployment/conf/.templates/setenv.sh.templ
MrCreosote: :+1:

.github/workflows/test.yml
MrCreosote: Nice! :+1:

.gitignore
MrCreosote: :+1:

.github/workflows/test.yml
MrCreosote: There are now docs changes in this PR - are there any more to do?

build.xml
MrCreosote: Uh oh, looks like the changes to the build.xml file got nuked

Dockerfile
MrCreosote: :+1:

deployment/conf/.templates/deployment.cfg.templ
MrCreosote: :+1:

deployment/conf/.templates/setenv.sh.templ
MrCreosote: :+1:

scripts/run_tests.sh
MrCreosote: could be part of the test startup actually

Dockerfile
ialarmedalien: Yes -- those folders aren't used as tomcat does the server work on this image. Other KBase java containers -- e.g. the auth2 image -- use jetty as a server so this sets up the appropriate directories for them.

docker-compose.yml
ialarmedalien: I've added some instructions to the README on how to pull the latest docker images from the GitHub container repo.

scripts/workspace_container_test.py
ialarmedalien: There's some fiddling required due to the location of the `biokbase` modules and this script. I've changed this so it now recommends running the wrapper script instead, which sidesteps all that nastiness.

scripts/run_tests.sh
ialarmedalien: I was having trouble concocting the appropriate bash invocation to wait for the services to be ready, but I've cracked it now. 🎉

.github/workflows/test.yml
ialarmedalien: Discussed over slack; see also #667 .

build.xml
ialarmedalien: they should be back in their full glory now.

.github/workflows/test.yml
MrCreosote: :+1:

scripts/workspace_container_test.py
MrCreosote: :+1:

build.xml
MrCreosote: :+1:

.github/workflows/test.yml
MrCreosote: :+1:

docker-compose.yml
MrCreosote: :+1:

readme.md
MrCreosote: :+1:

scripts/run_tests.sh
MrCreosote: :+1:

scripts/run_tests.sh
MrCreosote: :+1:

Dockerfile
MrCreosote: :+1:

scripts/run_tests.sh
MrCreosote: although I would probably still do it in the tests so they're more self contained, but this is fine

Dockerfile
MrCreosote: My particular concern is that it's not clear what version of java is being run and whether that will change as the ubuntu image is updated (my understanding it that it's not supposed to but...). Are you sure that java 11 is what's being run here?

If so and you're going to update the image to be more explicit about the java version being run I think that's ok

scripts/workspace_container_test.py
```diff
@@ -0,0 +1,76 @@
+from .lib.biokbase.workspace.client import Workspace
```
MrCreosote, line 1: In an ideal world we'd write this in java but this is fine

MrCreosote (changes requested):
Oops, looks like the container tests are failing but being marked as passing, that's not great
https://github.com/kbase/workspace_deluxe/actions/runs/4126618976/jobs/7128784932#step:5:3013

Dockerfile
ialarmedalien:
```sh
❯ docker build . -t workspace_image
[...]
❯ docker run -it workspace_image bash
```
```bash
root@aeb6bbc85848:/kb/deployment/services/workspace/tomcat# which java 
/usr/bin/java
root@aeb6bbc85848:/kb/deployment/services/workspace/tomcat# java --version
openjdk 11.0.17 2022-10-18
OpenJDK Runtime Environment (build 11.0.17+8-post-Ubuntu-1ubuntu218.04)
OpenJDK 64-Bit Server VM (build 11.0.17+8-post-Ubuntu-1ubuntu218.04, mixed mode, sharing)
root@aeb6bbc85848:/kb/deployment/services/workspace/tomcat# update-alternatives --config java
There is only one alternative in link group java (providing /usr/bin/java): /usr/lib/jvm/java-11-openjdk-amd64/bin/java
Nothing to configure.
```

ialarmedalien: ...and I do plan to update the image to be explicit about the version of java installed.

scripts/run_tests.sh
```diff
@@ -0,0 +1,26 @@
...
+max_retry=30
+counter=0
+
+# - wait for the workspace container to start up -- logs should stop being generated and the
+# last line of the logs should be something like
+# <timestamp> INFO [main] org.apache.catalina.startup.Catalina.start Server startup in 3811 ms
+until docker logs -n 5 workspace_deluxe-workspace-1 2>&1 | grep -q -F -m 1 'org.apache.catalina.startup.Catalina.start Server startup in' > /dev/null
+do
+    sleep 2
+    [[ counter -eq $max_retry ]] && echo "Failed!" && exit 1
```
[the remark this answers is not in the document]
ialarmedalien: Added in some log lines to help out here

scripts/run_tests.sh
```diff
@@ -0,0 +1,54 @@
...
+    fi
+    echo "Waiting for the workspace to be ready..."
+    counter=$(( counter + 1))
+    sleep 2
+done
+
+if [ $counter -eq $max_retries ]; then
+    echo "Workspace server start up not detected after $max_retries retries. "
+    echo "Last 20 lines of workspace container logs:"
+    docker logs -n 20 "workspace_deluxe-workspace-1"
```
ialarmedalien, line 49: example failure: https://github.com/kbase/workspace_deluxe/actions/runs/4135252550/jobs/7147454215#step:5:3010

scripts/run_tests.sh
```diff
@@ -0,0 +1,54 @@
...
+# limit the number of retries of the script
+while [ $counter -lt $max_retries ]; do
+    # wait for the workspace container to start up
+    # logs should stop being generated and the last line of the logs should be something like
+    # <timestamp> INFO [main] org.apache.catalina.startup.Catalina.start Server startup in 3811 ms
+    if docker logs -n 5 "workspace_deluxe-workspace-1" 2>&1 | grep -q -F -m 1 'org.apache.catalina.startup.Catalina.start Server startup in' > /dev/null; then
+        # get the path to the 'scripts' directory and add it to the python execution path
+        current_dir="$( dirname -- "$( readlink -f -- "$0"; )"; )"
+        PYTHONPATH="$current_dir":$PYTHONPATH python -m pytest scripts/workspace_container_test.py
+        exit_code=$?
```
ialarmedalien, line 38: example python test failure: https://github.com/kbase/workspace_deluxe/actions/runs/4135234730/jobs/7147411009#step:5:5425

scripts/run_tests.sh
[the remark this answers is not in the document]
MrCreosote: :+1:

MrCreosote (approved):
This is great, thanks for doing this

PR #664, opened 2023-02-02
CE-99: Credit metadata class

ialarmedalien:
structure and info from the spec:

> Currently the KBase workspace only supports credit metadata for objects of type 'dataset'; anything else will return an error.
> 
> The license may be supplied either as an URL pointing to licensing information for the resource, or using a license name. See https://choosealicense.com/appendix/ for a list of common open source licenses.
>
> Required fields are:
> - identifier
> - resource_type
> - versioning information: if the resource does not have an explicit version number, one or more dates should be supplied: ideally the date of resource publication and the last update (if applicable).
> - contributors (one or more required)
> - titles (one or more required)

```
	typedef structure {
		string identifier;
		string license;
		string resource_type;
		string version;
		list<String> comments;
		list<Contributor> contributors;
		list<EventDate> dates;
		list<FundingReference> funding;
		list<PermanentID> related_identifiers;
		list<Title> titles;
	} CreditMetadata;
```


src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,346 @@
...
+						license = null;
+					}
+				}
+				catch (Exception e) {
+					errorList.add(e.getMessage());
+				}
+			}
+
+
+			if (errorList.isEmpty()) {
```
ialarmedalien, line 335: TODO: add in check for having at least one contributor, but I wanted to quickly discuss it with Elisha first.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,346 @@
...
+		 */
+		public CreditMetadata build() {
+
+			if (version == null && (dates == null || dates.isEmpty())) {
+				errorList.add("must provide either 'version' or one or more 'dates', ideally indicating when the resource was published or when it was last updated");
+			}
+
+			// If the license looks like a URL, ensure it's properly formatted and seems valid.
+			// For these purposes, if it starts with http, assume it's an URL.
+			if (license != null && license.toLowerCase().startsWith("http")) {
```
ialarmedalien, line 319: This is a fairly basic (**cough**crap**cough**) way to detect an URL-like string, but should be OK as long as there are instructions to put in the fully-specified URL and examples of such data.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,346 @@
...
+				try {
+					final URL licenseURL = Common.processURL(license, "license");
+					if (licenseURL != null) {
+						license = licenseURL.toString();
```
ialarmedalien, line 323: The field itself is a string so once the URL has been standardised, it can be converted back to string form.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,655 @@
...
+	}
+
+	@Test
+	public void buildCleanURLTypeLicense() throws Exception {
+		for (final ResourceType rt : ResourceType.values()) {
+			final CreditMetadata creditMetadata = CreditMetadata
+					.getBuilder(IDENTIFIER_STRING, rt, TITLE_LIST)
+					.withLicense(LICENSE_URL_STRING)
+					.withVersion(VERSION_STRING_UNTRIMMED)
+					.build();
```
ialarmedalien, line 320: TODO: add a couple more examples here using `VALID_URL_LIST` in `[src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java](https://github.com/kbase/workspace_deluxe/pull/664/files#diff-f2c1dbae76bd1b1341255d67aca320f1e4d1c0c4fad611e3566a02dfb8b680d1)`.

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
```diff
@@ -25,6 +26,8 @@ public class ProvenanceTestCommon {
 		" \f ",
 		"\r",
 		"\n",
+
```
ialarmedalien, line 29: TODO: remove this line

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
```diff
@@ -59,4 +62,99 @@ public class ProvenanceTestCommon {
 		validPids.put(" https://this.is.the.url/ ", "https://this.is.the.url/");
 		VALID_PID_MAP = Collections.unmodifiableMap(validPids);
 	}
+
+	public static final List<String> VALID_URL_LIST = Collections.unmodifiableList(Arrays.asList(
```
ialarmedalien, line 66: TODO: use some of these as examples in the license URL tests

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
```diff
@@ -59,4 +62,99 @@ public class ProvenanceTestCommon {
...
+		"http://例子.测试",
+		"http://-.~_!$&'()*+,;=:%40:80%2f::::::@example.com",
+		"http://1337.net",
+		"http://a.b-c.de",
+		"http://223.255.255.254"
+	));
+
+	public static final Map<String, String> INVALID_URL_MAP;
+	static {
+		Map<String, String> invalidUrlMap = new HashMap<>();
```
ialarmedalien, line 104: TODO: sort this set of dodgy URLs out and winnow it down to create a reasonable test set.

Unfortunately the source of these sample invalid URLs used a slightly different test for URL validity to that used by java, so some of these supposedly invalid URLs are considered OK by java.

There is also an issue with incomplete URLs of the form `http://<special characters here>` where java removes the `//` from the URL string, which makes it more difficult to match up error strings.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
ialarmedalien: DONE!

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
ialarmedalien: DONE!

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
ialarmedalien: DONE!

(Decided not to fix the underlying issues with the Java URL library... selfish, I know)

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: DONE!

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,349 @@
...
+		 * @return the credit metadata.
+		 */
+		public CreditMetadata build() {
+
+			if (version == null && (dates == null || dates.isEmpty())) {
+				errorList.add("must provide either 'version' or one or more 'dates', ideally indicating when the resource was published or when it was last updated");
+			}
+
+			// If the license looks like a URL, ensure it's properly formatted and seems valid.
+			// For these purposes, if it starts with http or contains '://', assume it's an URL.
```
ialarmedalien, line 321: This is a fairly basic (cough_crap_cough) way to detect an URL-like string, but should be OK as long as there are instructions to put in the fully-specified URL and examples of such data.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: moved comment to appropriate line in most recent commit

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
ialarmedalien: DONE!

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+		private String identifier;
+		private String license = null;
+		private String version = null;
+		private ResourceType resourceType = ResourceType.DATASET;
+		private List<Contributor> contributors;
+		private List<EventDate> dates = null;
+		private List<FundingReference> funding = null;
+		private List<PermanentID> relatedIdentifiers = null;
+		private List<Title> titles;
+		private List<String> errorList;
```
MrCreosote, line 211: Why are some fields initialized with null while other's aren't?

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+				errorList.add(errorMessage);
+			}
+		}
+		return new Builder(identifier, rt, contributors, titles, errorList);
+	}
+
+	/** A builder for {@link CreditMetadata}. */
+	public static class Builder {
+		private String comments;
+		private String identifier;
```
MrCreosote, line 202: Mark required fields as final

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+		 */
+		public CreditMetadata build() {
+
+			if (version == null && (dates == null || dates.isEmpty())) {
+				errorList.add("must provide either 'version' or one or more 'dates', ideally indicating when the resource was published or when it was last updated");
+			}
+
+			// If the license looks like a URL, ensure it's properly formatted and seems valid.
+			// For these purposes, if it starts with http or contains '://', assume it's an URL.
+			if (license != null && (license.toLowerCase().startsWith("http") || license.contains("://"))) {
```
MrCreosote, line 322: Why not just run processUrl if the string isn't null and check the protocol is http/s, and treat it as text otherwise?

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+				try {
+					final URL licenseURL = Common.processURL(license, "license");
+					// license field is a string, so convert the URL back to string form
+					license = licenseURL.toURI().normalize().toString();
```
MrCreosote, line 326: toURI can throw an exception that probably isn't going to be very useful for users debugging-wsie

MrCreosote: Just as a thought, you may want to write some integration tests at the argutils level when coverting the generated classes to these classes to see how hard it is to debug an error deep in the class hierarcy

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+				&& Objects.equals(version, other.version);
+	}
+
+	/**
+	 * Gets a builder for the {@link CreditMetadata}.
+	 *
+	 * @param identifier unique persistent ID for the resource
+	 * @param resourceType type of the resource, as a {@link ResourceType}
+	 * @param contributors list of {@link Contributor} objects
+	 * @param titles list of {@link Title} objects
```
MrCreosote, line 152: Should mention that at least one title / contributor is required, here and below

MrCreosote (commented):
The comments here might affect the tests, so I'll review them after these are sorted

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -34,7 +35,27 @@ static String processString(final String input) {
...
+		}
+		final String[] lines = input.split("\n");
+		final String joined = Arrays.stream(lines)
+			.map(String::trim)
+			.filter(item -> !item.isEmpty())
+			.collect(Collectors.joining("\n"));
+		return processString(joined);
+	}
+
+	public static URL processURL(final String url, final String name) {
```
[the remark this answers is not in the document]
ialarmedalien: ah crap, I made it public whilst testing and forgot to change it back

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+	 * @return the builder.
+	 */
+	public static Builder getBuilder(
+			final String identifier,
+			final ResourceType resourceType,
```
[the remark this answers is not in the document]
ialarmedalien: yeah, wasn't sure how to handle this since I think it's the only place where this is being done

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: how is that advantageous over what is written here?

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: `processUrl` wraps the error in a comforting message saying "Illegal license URL <url>: <specific error>", so they should at least realise that the URL is malformed.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: The field that wasn't initialised was added later and I obviously forgot to set it to null.

ialarmedalien:
Force-pushed over a commit that included whitespace changes and content changes to split them out into separate commits for ease of reviewing.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
ialarmedalien: Current mode of operation is to use the default if the input is null, and if there is input (including the empty string / whitespace only strings), check that it is valid. That will catch errors if some other value is put in the field but allows input to be null to save vital keystrokes for tired users.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: Oh, toURI() is already called there, ok.

:+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: Slightly simpler code - no need for the "does this look vaguely like a URL" check, just have processURL do it for you.

Also then you can do the equivalent of
```
if (!new HashSet<>(Arrays.asList("http", "https")).contains(url.getProtocol())) {
    // handle error
}
```

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: But if there's a default you don't need to have any input... that's the point of a default...?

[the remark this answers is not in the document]
ialarmedalien: The spec says:

> The 'resource_type' field should be filled using values from the DataCite resourceTypeGeneral field:
> 
> 		https://support.datacite.org/docs/datacite-metadata-schema-v44-mandatory-properties#10a-resourcetypegeneral
> 
> 		Currently the KBase workspace only supports credit metadata for objects of type
> 		'dataset'; anything else will return an error.
>
> [...]
> resource_type - the broad type of the source data for this workspace object.
> 			Defaults to 'dataset' (the only valid value currently) if no value is
> 			provided.
> 			Valid 'resource_type' values:
> 				- dataset

Based on that, the expectation for incoming `resource_type` fields would be
- `resource_type` field absent
- `resource_type` field present, value null
- `resource_type` field present, contains a string value

In ArgUtils, when converting incoming `us.kbase.workspace.CreditMetadata` into `...database.provenance.CreditMetadata`, do something like:
```java
CreditMetadata.getBuilder(
        cm.getIdentifier(),
        cm.getResourceType(),   // or try/catch if cm.getResourceType() throws an NPE
        cm.getContributors(),
        cm.getTitles())
        [...]
```

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: Are you proposing something like this?

```java

if (license != null) {
	try {
		final URL licenseURL = Common.processURL(license, "license");
		// license field is a string, so convert the URL back to string form
		license = licenseURL.toURI().normalize().toString();
		if (!new HashSet<>(Arrays.asList("http", "https")).contains(licenseURL.getProtocol())) {
			// handle error
		}
	} catch (Exception e) {
		// treat as a string
		license = license.trim();
	}

```
If I'm understanding your suggestion right, that method doesn't differentiate between an invalid URL and a text string, though.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: But you could do the equivalent with
```java
CreditMetadata.getBuilder(
        cm.getIdentifier(),
        cm.getContributors(),
        cm.getTitles())
        .withResourceType(cm.getResourceType()
        ...
```

And then users of the class don't have to stick in an extra, unnecessary `null` argument when they want to use the default.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: Ok, I see, you want the `looks vaguely like a url check` to cause the method to reject things that aren't urls but look like urls.

It seems unlikely that license text starts with HTTP[S], but it seems pretty likely that there might be another url within the text, so I'm not sure how helpful the `://` check is. Maybe just check it starts with `http://` or `https://`?

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+		this.funding = funding;
+		this.identifier = identifier;
+		this.license = license;
+		this.relatedIdentifiers = relatedIdentifiers;
+		this.resourceType = resourceType;
+		this.titles = titles;
+		this.version = version;
+	}
+
+	public String getIdentifier() {
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+	public List<Title> getTitles() {
+		return Common.getList(titles);
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(comments, contributors, dates, funding, identifier, license, relatedIdentifiers,
+				resourceType,
+				titles,
+				version);
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,345 @@
...
+
+		/**
+		 * Build the {@link CreditMetadata}.
+		 *
+		 * @return the credit metadata.
+		 */
+		public CreditMetadata build() {
+
+			if (version == null && (dates == null || dates.isEmpty())) {
+				errorList.add("must provide either 'version' or one or more 'dates', ideally indicating when the resource was published or when it was last updated");
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: `identifier` should be required right?

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: Yes, but it needs to be massaged with `checkPid` first ==> it isn't set as `final`

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
ialarmedalien: ah OK, that looks reasonable. Will implement.

MrCreosote (commented):
Holding off on tests for now as there may be significant restructuring of the class API

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: I don't think I understand - you can do whatever you want to the data you're assigning to a final instance variable before you assign it, you just can't assign it more than once. Looking at the builder constructor it looks to me like making the instance variable `final` would be fine

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,440 @@
...
+	 * @return list of {@link Contributor} objects.
+	 */
+	public List<Contributor> getContributors() {
+		return Common.getList(contributors);
+	}
+
+	/*
+	 * Gets any lifecycle dates.
+	 *
+	 * @return list of {@link EventDate} objects (if present)
```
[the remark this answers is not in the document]
ialarmedalien: deleted the errant `(if present)`

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,440 @@
...
+		 * Build the {@link CreditMetadata}.
+		 *
+		 * @return the credit metadata.
+		 */
+		public CreditMetadata build() {
+
+			if (version == null && (dates == null || dates.isEmpty())) {
+				errorList.add("must provide either 'version' or one or more " +
+				"'dates', ideally indicating when the resource was published or " +
+				"when it was last updated");
```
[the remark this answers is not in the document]
ialarmedalien: fixed

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,440 @@
...
+	 * Gets the comments for the resource.
+	 *
+	 * @return comments (if present)
+	 */
+	public Optional<String> getComments() {
+		return Optional.ofNullable(comments);
+	}
+
+	/*
+	 * Gets the license for the resource.
```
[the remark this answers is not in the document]
ialarmedalien: documented

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -34,6 +35,26 @@ static String processString(final String input) {
 		return isNullOrWhitespace(input) ? null : input.trim();
 	}
 
+	/**
+	 * Trim strings and remove empty lines in a multi-line string
+	 * @param input
+	 *                string to be parsed as a multi-line string
+	 * @return
+	 *                string comprised of the input lines, joined with "\n",
+	 *                with empty and whitespace-only lines removed
```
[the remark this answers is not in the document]
ialarmedalien: I have changed the field to be a list of strings, so it's a moot point.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,440 @@
...
+		/**
+		 * Build the {@link CreditMetadata}.
+		 *
+		 * @return the credit metadata.
+		 */
+		public CreditMetadata build() {
+
+			if (version == null && (dates == null || dates.isEmpty())) {
+				errorList.add("must provide either 'version' or one or more " +
+				"'dates', ideally indicating when the resource was published or " +
```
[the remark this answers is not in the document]
ialarmedalien: I will leave it the way it is.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: New set up: if it matches `^[a-zA-Z]+:+/+`, treat it like an URL. Otherwise, leave as-is. Covers typos in the protocol and colon-slash part (or valid alternate protocols like ftp), and the intelligent user will be able to deduce that multiple URLs can be entered by preceding them with some textual filler that doesn't match the afore-mentioned regex.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
ialarmedalien: Once this class and the container class are done, I was planning to add a short description to the class doctoring to summarise required/optional fields and check it all tallies with the spec and the original linkml schema.

src/us/kbase/workspace/database/provenance/Common.java
```diff
@@ -34,7 +35,27 @@ static String processString(final String input) {
...
+	 */
+	static String processMultilineString(final String input) {
+		if (input == null) {
+			return null;
+		}
+		final String[] lines = input.split("\n");
+		final String joined = Arrays.stream(lines)
+			.map(String::trim)
```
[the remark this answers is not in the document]
MrCreosote: This code was removed so this comment is no longer relevant

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: I'm a bit iffy about this but not enough to actually do anything other than comment I'm iffy, so

:+1:

src/us/kbase/workspace/database/provenance/Common.java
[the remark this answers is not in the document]
MrCreosote: moooot :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: Looks like the requirement is documented in the builder docstring, so that's good enough for me

:+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+	 * @return resource version (if present)
+	 */
+	public Optional<String> getVersion() {
+		return Optional.ofNullable(version);
+	}
+
+	/*
+	 * Gets the resource type.
+	 *
+	 * @return resource type as a {@link ResourceType}
```
MrCreosote, line 113: This is totally fine, but for future reference the java docs will automatically link to the definition of the return type so the link is redundant

ialarmedalien: So what should I put in the future for this kind of thing?

```java
 *
 * @return the resource type as a ResourceType
 */
 ```
 
 or just
 ```java
 *
 * @return the resource type
 */

File: src/us/kbase/workspace/database/provenance/CreditMetadata.java [hunk 26]
```diff
@@ -0,0 +1,465 @@
...
+				&& Objects.equals(resourceType, other.resourceType)
+				&& Objects.equals(titles, other.titles)
+				&& Objects.equals(version, other.version);
+	}
+
+	/**
+	 * Gets a builder for the {@link CreditMetadata}.
+	 *
+	 * @param identifier
+	 *                unique persistent ID for the resource
```
MrCreosote, line 212: I've realized I have no idea what the `resource` is in this context or what the ID should look like, if anything. Can you add some docs around that?

Edit: looks like it needs to pass the `checkPID` func.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+			if (this.titles == null || this.titles.isEmpty()) {
+				this.errorList.add("at least one title must be provided");
+			}
+		}
+
+		/**
+		 * Sets the resource type for the resource.
+		 *
+		 * As there is currently only one resource type, entering null will set the
+		 * resource type value to DEFAULT_RESOURCE_TYPE in the resultant object.
```
MrCreosote, line 270:
```suggestion
		 * Entering null will set the resource type value to DEFAULT_RESOURCE_TYPE
		 * in the resultant object.
```
I don't think the language around there being a single resource type is necessary, it's totally fine to have an input that in the future may be expanded to more choices and is currently automatically set to what in the future will be the default

MrCreosote, line 270: Alternatively, you could just ignore null / whitespace only input as a noop rather than resetting to the default. In the current case, there would be no behavior change

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+		 * @param resourceType
+		 *                resource type as a string
+		 * @return this builder
+		 */
+		public Builder withResourceTypeString(final String resourceType) {
+			if (resourceType != null) {
+				try {
+					this.resourceType = ResourceType.getResourceType(resourceType);
+				} catch (IllegalArgumentException e) {
+					// TEMPORARY HACK
```
MrCreosote, line 281: I don't think this behavior will change when there is more than one resource type, right? whitespace will still be an error and the message will still need editing, and null should still default to `DEFAULT_RESOURCE_TYPE`

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+		 *
+		 * As there is currently only one resource type, entering null will set the
+		 * resource type value to DEFAULT_RESOURCE_TYPE in the resultant object.
+		 *
+		 * @param resourceType
+		 *                resource type as a {@link ResourceType}
+		 * @return this builder
+		 */
+		public Builder withResourceType(final ResourceType resourceType) {
+			// TEMPORARY HACK - only one resource type at present
```
MrCreosote, line 313: Again, I don't think this is a temporary hack, it's perfectly fine long term. The only thing I'd consider is ignoring null rather than resetting to default, but I'm not religious there.

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+			else {
+				this.resourceType = DEFAULT_RESOURCE_TYPE;
+			}
+			return this;
+		}
+
+		/**
+		 * Sets the resource type for the resource.
+		 *
+		 * As there is currently only one resource type, entering null will set the
```
MrCreosote, line 305: Same as above re language

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+		 * Sets the version.
+		 *
+		 * Valid credit metadata objects require either a version or one or more dates.
+		 *
+		 * @param withVersion
+		 *                version string
+		 * @return this builder
+		 */
+		public Builder withVersion(final String version) {
+			this.version = Common.processString(version);
```
MrCreosote, line 366: Do you want to have a maximum length for the version (and I guess any of these fields)? We'll definitely need a check for the overall size of the CM regardless

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+		 * requires some sort of versioning information -- either an explicit version
+		 * identifier or one or more dates.
+		 *
+		 * @return the credit metadata.
+		 */
+		public CreditMetadata build() {
+
+			// ensure there is either a version or a non-empty list of dates
+			if (version == null && (dates == null || dates.isEmpty())) {
+				errorList.add("must provide either 'version' or one or more " +
```
MrCreosote, line 421: Do you want to check for a specific set of event types?

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+						"'dates', ideally indicating when the resource " +
+						"was published or when it was last updated");
+			}
+
+			// If the license looks like a URL, ensure it's properly formatted and seems
+			// valid. For these purposes, a license string that starts with a series
+			// of letters followed by one or more colons and one or more slashes is
+			// considered URL-like.
+			Pattern pattern = Pattern.compile("^[a-zA-Z]+:+/+");
```
MrCreosote, line 430: Make this static so it's only compiled once

MrCreosote (changes requested):
Doing tests in a separate review

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: The latter is what I do. The getters and setters we're dealing with in a value class are usually so simple I hardly thing javadocs are necessary, but it's considered bad form so...

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: Oh, you might want to make `DEFAULT_RESOURCE_TYPE` a link come to think of it

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+			for (final String licenseURL : VALID_URL_LIST) {
+				final CreditMetadata creditMetadata2 = CreditMetadata
+						.getBuilder(IDENTIFIER_STRING, CONTRIBUTOR_LIST, TITLE_LIST)
+						.withResourceType(rt)
+						.withLicense(licenseURL)
+						.withDates(DATE_LIST)
+						.build();
+				assertCreditMetadataFields(creditMetadata2, IDENTIFIER_STRING, rt,
+						CONTRIBUTOR_LIST, TITLE_LIST,
+						ImmutableMap.of(LICENSE, licenseURL),
```
MrCreosote, line 356: The build step might mutate the url by normalizing it - currently that's untested AFAICT

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+					.withFunding(FUNDING_LIST)
+					.withFunding(null)
+					.withRelatedIdentifiers(RELATED_ID_LIST)
+					.withRelatedIdentifiers(ELRI)
+					.withVersion(VERSION_STRING_UNTRIMMED)
+					.build();
+			assertCreditMetadataFields(creditMetadata, IDENTIFIER_STRING, rt,
+					CONTRIBUTOR_LIST, TITLE_LIST, VERSION_MAP);
+		}
+	}
```
MrCreosote, line 555: Are there any tests removing nulls & duplicates from the non-string lists?

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+			fail(EXP_EXC);
+		} catch (Exception got) {
+			TestCommon.assertExceptionCorrect(got, new IllegalArgumentException(
+					"Errors in CreditMetadata construction:\n" + errorString));
+		}
+	}
+
+	@Test
+	public void buildFailInvalidPID() throws Exception {
+		for (final ResourceType rt : ResourceType.values()) {
```
MrCreosote, line 578: Again, if the purpose of the test and tests below is to check a specific error, I don't understand the benefit of cycling through the resource types

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+							.withDates(Arrays.asList(null, null, null)),
+					VERSIONING_DATES_REQUIRED);
+
+		}
+	}
+
+	@Test
+	public void buildFailNullOrEmptyContributorList() throws Exception {
+		for (final ResourceType rt : ResourceType.values()) {
+			// null title list
```
MrCreosote, line 759: test says contributor list, this says title list

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+							CONTRIBUTOR_LIST,
+							Arrays.asList(null, null, null, null))
+							.withResourceType(rt)
+							.withVersion(VERSION_STRING),
+					TITLE_AT_LEAST_ONE);
+		}
+	}
+
+	@Test
+	public void buildMinimalFailAllFields() throws Exception {
```
MrCreosote, line 815: Do you want to test a bad license url here?

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: Oops, the top one doesn't trim the strings. This should work:
```java
.filter(c -> c != null && !c::isBlank())
.map(String::strip)
```

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,465 @@
...
+		 * @param comments
+		 *                comments as a list of strings
+		 * @return this builder
+		 */
+		public Builder withComments(final List<String> comments) {
+			if (comments != null) {
+				this.comments = comments.stream()
+						.filter(Objects::nonNull)
+						.map(String::trim)
+						.filter(c -> !Util.isNullOrWhitespace(c))
```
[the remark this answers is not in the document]
MrCreosote: Oops, the top one doesn't trim the strings. This should work:
```java
.filter(c -> c != null && !c::isBlank())
.map(String::strip)
```

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+			"data sets",
+			TYPE,
+			"DATAZET"
+	};
+
+	private static final String TITLE_STRING = "A Series of Unfortunate Elephants";
+
+	private static final List<String> COMMENTS_LIST = Arrays.asList(
+			"An Airport For Aliens Currently Run By Dogs",
+			"Space Warlord Organ Trading Simulator",
```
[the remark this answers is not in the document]
ialarmedalien: https://store.steampowered.com/app/1507780/Space_Warlord_Organ_Trading_Simulator/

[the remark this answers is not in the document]
MrCreosote: Wow. I should've expected that I suppose

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
```diff
@@ -59,4 +61,66 @@ public class ProvenanceTestCommon {
...
+	public static final List<String> VALID_URL_LIST = Collections.unmodifiableList(Arrays.asList(
+			"http://foo.com/blah_blah",
+			"http://foo.com/blah_blah/",
+			"http://foo.com/blah_blah_(wikipedia)",
+			"http://foo.com/blah_blah_(wikipedia)_(again)",
+			"http://www.example.com/wpstyle/?p=364",
+			"https://www.example.com/foo/?bar=baz&inga=42&quux",
+			"http://✪df.ws/123",
+			"http://userid:password@example.com:8080",
+			"http://userid:password@example.com:8080/",
```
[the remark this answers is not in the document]
ialarmedalien: I would very much doubt that situation would occur -- this was just a list of URLs I found to be used for testing an URL validator. I had to remove quite a few of the list as Java's URL / URI classes happily passed a number of invalid strings.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+	 *                list of PermanentIDs expected to be returned by
+	 *                `getRelatedIdentifiers`;
+	 *                if null, should return an empty list
+	 */
+	private void assertCreditMetadataFields(final CreditMetadata creditMetadata,
+			final String identifier,
+			final ResourceType rt,
+			final List<Contributor> contributors,
+			final List<Title> titles,
+			final Map<String, String> expectedMap,
```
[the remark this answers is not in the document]
ialarmedalien: There were three args so it made more sense then. I just didn't change it when I removed the other arg.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+		}
+	}
+
+	// Note: since there is currently only one ResourceType, not supplying a value
+	// for the field currently results in the default (only) value being used.
+	// This behaviour will change when more resource types are added, and these
+	// tests
+	// will need to be updated.
+	@Test
+	public void buildMinimalDefaultResourceType() throws Exception {
```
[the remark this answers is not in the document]
ialarmedalien: There is now

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: No longer applicable

ialarmedalien: N/A

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: N/A

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: N/A

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: N/A

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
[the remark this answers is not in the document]
ialarmedalien: Moved these examples to a separate set of tests that ensure that an error is thrown if the URL contains user/password info.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+				assertCreditMetadataFields(creditMetadata, entry.getValue(),
+						rtes.getValue(), CONTRIBUTOR_LIST, TITLE_LIST,
+						VERSION_MAP);
+			}
+		}
+	}
+
+	// Note: since there is currently only one ResourceType, not supplying a value
+	// for the field currently results in the default (only) value being used.
+	// This behaviour will change when more resource types are added, and these
```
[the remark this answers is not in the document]
ialarmedalien: N/A

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+
+	@Test
+	public void buildMinimal() throws Exception {
+		for (final ResourceType rt : ResourceType.values()) {
+			for (Map.Entry<String, String> entry : VALID_PID_MAP.entrySet()) {
+				final CreditMetadata creditMetadata = CreditMetadata
+						.getBuilder(entry.getKey(), CONTRIBUTOR_LIST,
+								TITLE_LIST)
+						.withVersion(VERSION_STRING)
+						.withResourceType(rt)
```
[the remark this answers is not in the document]
ialarmedalien: N/A

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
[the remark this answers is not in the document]
ialarmedalien: N/A

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,883 @@
...
+				.build();
+
+		assertCreditMetadataFields(creditMetadata2, IDENTIFIER_STRING, ResourceType.DATASET,
+				CONTRIBUTOR_LIST, TITLE_LIST, VERSION_MAP);
+
+	}
+
+	@Test
+	public void buildCleanAndTrimComments() throws Exception {
+		for (final ResourceType rt : ResourceType.values()) {
```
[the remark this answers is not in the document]
ialarmedalien: N/A - changed all these tests

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
ialarmedalien: N/A

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: I will retrofit sizes if it seems like it's something we should add.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
ialarmedalien: see the test `buildCleanURLTypeLicense`

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: Could you note in the docstring that  null, blank, and duplicate strings are filtered out?

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: I see it's now final, but it'll still be compiled every call - I mean to make it a static class variable so it's only compiled once the first time the class is imported

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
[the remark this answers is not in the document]
MrCreosote: So do you still not want to change it? IMO it'd be better without the map

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/ProvenanceTestCommon.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: The input to the CM and the input to the CM contents checker function are both `licenseURL`, which means the URL wasn't normalized / mutated, right? I'm not understanding how the normalization step is tested

MrCreosote: To further clarify, if the `.normalize()` step in converting the URLs to strings was removed from the CM class, all the tests would pass AFAICT. If the normalize is important there should be at least one test that would fail

MrCreosote: To be even more pedantic and annoying https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#normalize--

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,919 @@
...
+
+	@Test
+	public void buildFailMissingVersionOrDates() throws Exception {
+		// no version or dates at all
+		buildCreditMetadataFailWithError(
+				CreditMetadata.getBuilder(
+						IDENTIFIER_STRING,
+						RESOURCE_TYPE_STRING,
+						CONTRIBUTOR_LIST,
+						TITLE_LIST),				VERSIONING_DATES_REQUIRED);
```
MrCreosote, line 656: Odd whitespace on this line

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,466 @@
...
+	/**
+	 * Gets a builder for the {@link CreditMetadata}.
+	 *
+	 * @param identifier
+	 *                unique persistent ID for the resource (i.e. the source data for this
+	 *                workspace object). Should be in the format
+	 *                <database name>:<identifier within database>
+	 * @param resourceType
+	 *                type of the resource, as a {@link ResourceType}
+	 * @param contributors
```
MrCreosote, line 215: This also needs no changes, but inputs are also automatically linked by javadoc so manual links aren't necessary, although they don't hurt
. The link to the CreditMetadata class *is* helpful though.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,919 @@
...
+				.withComments(COMMENTS_LIST_UNTRIMMED)
+				.withDates(DATE_LIST_DUPES_NULLS)
+				.withFunding(FUNDING_LIST_DUPES_NULLS)
+				.withRelatedIdentifiers(RELATED_ID_LIST_DUPES_NULLS)
+				.build();
+				assertCreditMetadataFields(creditMetadata, IDENTIFIER_STRING, RESOURCE_TYPE,
+				CONTRIBUTOR_LIST, TITLE_LIST, EM,
+				COMMENTS_LIST, DATE_LIST, FUNDING_LIST,
+				RELATED_ID_LIST);
+	}
```
MrCreosote, line 518: Did you want to test for immutability of the CM class? Should have caught this one earlier...

MrCreosote (commented):
Just about done, just these and 5 or so open items above

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: Doh, got it.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
ialarmedalien: 😖

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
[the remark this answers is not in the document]
ialarmedalien: I don't want to change it, but... done.

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
ialarmedalien: Aren't all these classes going to need to be tested for immutability?

MrCreosote: I thought the ones that accepted or emitted mutable containers were tested? Did I miss some?

Either way, yeah if a value class accepts mutable containers or emits mutable containers testing that mutating the containers doesn't mutate the class should be done

src/us/kbase/workspace/database/provenance/CreditMetadata.java
ialarmedalien: Still thinking about this one.

ialarmedalien: I'm going to leave this unfiltered at the moment. There's only one event that might be dodgy and that one turns up very rarely in records (I haven't seen any instances of it from the bulk queries I've done).

MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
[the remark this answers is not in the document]
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: no changes needed :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,479 @@
...
+			}
+		}
+		return new Builder(identifier, rt, contributors, titles, errorList);
+	}
+
+
+	// Regular expression used to detect URL-like strings; detects
+	// a sequence of letters at the start of a line, followed by one or more
+	// colons and one or more slashes.
+	static final Pattern pattern = Pattern.compile("^[a-zA-Z]+:+/+");
```
MrCreosote, line 278: I would make this private just so the next person that comes along doesn't waste time trying to figure out why it isn't when it can be

src/us/kbase/workspace/database/provenance/CreditMetadata.java
```diff
@@ -0,0 +1,479 @@
...
+				errorList.add("must provide either 'version' or one or more " +
+						"'dates', ideally indicating when the resource " +
+						"was published or when it was last updated");
+			}
+
+			// If the license looks like a URL, ensure it's properly formatted and seems
+			// valid. For these purposes, a license string that starts with a series
+			// of letters followed by one or more colons and one or more slashes is
+			// considered URL-like.
+			final Pattern pattern = Pattern.compile("^[a-zA-Z]+:+/+");
```
MrCreosote, line 441: This line should be deleted I believe so the static variable is used

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,1052 @@
...
+		assertThat(INCORRECT_DATES, cm.getDates(),
+				is(dates));
+		assertThat(INCORRECT_FUNDING, cm.getFunding(),
+				is(funding));
+		assertThat(INCORRECT_RELATED_IDS, cm.getRelatedIdentifiers(),
+				is(relatedIdentifiers));
+		assertThat(INCORRECT_TITLES, cm.getTitles(),
+				is(titles));
+		assertThat(INCORRECT_LICENSE, cm.getLicense(), is(Optional.ofNullable(license)));
+		assertThat(INCORRECT_VERSION, cm.getVersion(), is(Optional.ofNullable(version)));
```
MrCreosote, line 279: These last two lines are recapitulating the code under test with the `ofNullable` stuff. Better to make the incoming type optional

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,1052 @@
...
+				RELATED_ID_LIST);
+	}
+
+	@Test
+	public void assertContributorsImmutable() throws Exception {
+		// same as CONTRIBUTOR_LIST
+		final List<Contributor> contributorList = new ArrayList<>(Arrays.asList(C1, C2));
+		final CreditMetadata cm = CreditMetadata
+				.getBuilder(IDENTIFIER_STRING, RESOURCE_TYPE_STRING,
+						CONTRIBUTOR_LIST, TITLE_LIST)
```
MrCreosote, line 507:
```suggestion
						contributorList, TITLE_LIST)
```

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,1052 @@
...
+	}
+
+	@Test
+	public void assertDatesImmutable() throws Exception {
+		// same as DATE_LIST
+		final List<EventDate> dateList = new ArrayList<>(Arrays.asList(ED1, ED2));
+		final CreditMetadata cm = CreditMetadata
+				.getBuilder(IDENTIFIER_STRING, RESOURCE_TYPE_STRING,
+						CONTRIBUTOR_LIST, TITLE_LIST)
+				.withDates(DATE_LIST)
```
MrCreosote, line 566:
```suggestion
				.withDates(dateList)
```

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1: but see https://github.com/kbase/workspace_deluxe/pull/664#discussion_r1113667872

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
ialarmedalien: Ugh... this is what happens when you try to code after waking up at ridiculous o'clock. (I know you're in a permanent state of sleep deprivation, but I'm not as good at it as you are.)

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
ialarmedalien: 🤦

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/database/provenance/CreditMetadata.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
```diff
@@ -0,0 +1,1049 @@
...
+
+import com.google.common.collect.ImmutableMap;
+import java.util.List;
+import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.Collections;
+import java.util.Optional;
+
+import static us.kbase.common.test.TestCommon.opt;
```
MrCreosote, line 20: I always put all static imports first just because that's how I've seen it done *shrug*

MrCreosote (approved):
LGTM assuming tests pass

src/us/kbase/workspace/test/database/provenance/CreditMetadataTest.java
ialarmedalien: I'll move them in the next PR.

PR #666, opened 2023-02-06
CE-99: use camelCase consistently for variable names containing 'ID' and 'URL'

ialarmedalien:
AKA the pull request of the beast

This PR converts the acronyms and abbreviations used in credit engine field names to use camel case, instead of a mixture of camel case and all caps. This promotes legibility and allows devs to apply consistent rules to method and attribute naming instead of having to wonder which acronyms do get all caps (ID? URL? HTTP? etc.) and which don't.

See https://google.github.io/styleguide/javaguide.html#s5.3-camel-case for the rules used.


scripts/run_tests.sh
```diff
@@ -40,7 +40,7 @@ while [ $counter -lt $max_retries ]; do
     fi
     echo "Waiting for the workspace to be ready..."
     counter=$(( counter + 1))
-    sleep 2
+    sleep 5
```
ialarmedalien, line 43: wait slightly longer during test start up in case GitHub is being slow

MrCreosote (approved):
LGTM

PR #671, opened 2023-02-09
CE-151: update to CreditMetadata schema

ialarmedalien:
Couple of minor spec changes:
- comments field is now a list of strings
- more text about the default resource_type, 'dataset'
- clarify license / license URL info
- rename CreditMetadataContainer -> CreditMetadataEntry to sound a bit more colloquial and less computer-y.

### DIFF
--- a/workspace.spec
+++ b/workspace.spec
@@ -963,14 +963,18 @@ module Workspace {
 
 		Required fields are:
 		- identifier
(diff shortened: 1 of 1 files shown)


MrCreosote (approved):
I assume this compiles, and if so, LGTM

PR #680, opened 2023-02-27
CE-151: Adding kb-sdk-generated classes for credit metadata

ialarmedalien:
Autogenerated classes for credit metadata, as generated by everyone's favourite compiler, kb-sdk.

### DIFF
--- a/src/us/kbase/workspace/Contributor.java
+++ b/src/us/kbase/workspace/Contributor.java
@@ -1,0 +1,235 @@
+
+package us.kbase.workspace;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Generated;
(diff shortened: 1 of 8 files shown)


MrCreosote (commented):
can you just run `make compile` so the other language libs, and more importantly, the html docs get updated?

Otherwise I'm assuming the changes came straight from the compiler and therefore don't need review

MrCreosote:
We may want to consider removing the generated code (except for the WorkspaceService) from the coverage calculations

ialarmedalien:
I transferred these over from the branch with the workspace spec updates in it just so I could work on the ArgUtils and other pieces that need these objects. When that other branch gets merged to develop, I can do a PR with all the docs updates.

MrCreosote (approved):
Fair enough

PR #684, opened 2023-04-11
CE-225: Award => grant, org name/ID to Organization object

ialarmedalien:
Some schema changes brought about after aligning the schema with schema.org:
- use an organisation object instead of having funder_name and funder_id
- switch from using 'award' to using 'grant' instead

I'll need to update the spec and the generated classes but have a few other schema alterations to make, so will do them first.

### DIFF
--- a/src/us/kbase/workspace/database/provenance/FundingReference.java
+++ b/src/us/kbase/workspace/database/provenance/FundingReference.java
@@ -98,7 +98,7 @@ public class FundingReference {
 	/** A builder for an {@link FundingReference}. */
 	public static class Builder {
 
-		private Organization funder;
+		private final Organization funder;
 		private String grantID = null;
 		private String grantTitle = null;
 		private URL grantURL = null;
--- a/src/us/kbase/workspace/database/provenance/FundingReference.java
+++ b/src/us/kbase/workspace/database/provenance/FundingReference.java
@@ -5,83 +5,69 @@ import java.util.Objects;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Optional;
-import us.kbase.workspace.database.Util;
 
 /**
- * A funding reference for a resource, including the funding body and the grant
- * awarded.
+ * A funding reference for a resource, including the funding body and the grant.
  */
 public class FundingReference {
 
-	private final String funderID;
-	private final String funderName;
-	private final String awardID;
-	private final String awardTitle;
-	private final URL awardURL;
+	private final Organization funder;
+	private final String grantID;
+	private final String grantTitle;
+	private final URL grantURL;
 
 	private FundingReference(
-			final String funderID,
-			final String funderName,
-			final String awardID,
-			final String awardTitle,
-			final URL awardURL) {
-		this.funderID = funderID;
-		this.funderName = funderName;
-		this.awardID = awardID;
-		this.awardTitle = awardTitle;
-		this.awardURL = awardURL;
+			final Organization funder,
+			final String grantID,
+			final String grantTitle,
+			final URL grantURL) {
+		this.funder = funder;
+		this.grantID = grantID;
+		this.grantTitle = grantTitle;
+		this.grantURL = grantURL;
 	}
 
 	/**
-	 * Gets the funder ID, for example ROR:04xm1d337.
+	 * Gets the funding organization.
 	 *
-	 * @return the funder ID, if present.
+	 * @return the funder organization.
 	 */
-	public Optional<String> getFunderID() {
-		return Optional.ofNullable(funderID);
+	public Organization getFunder() {
+		return funder;
 	}
 
 	/**
-	 * Gets the funder name, for example US Department of Energy.
-	 *
-	 * @return the funder name.
-	 */
-	public String getFunderName() {
-		return funderName;
-	}
-
-	/**
-	 * Gets the code assigned by the funder to the award, for example
+	 * Gets the code assigned by the funder to the grant, for example
 	 * DOI:10.46936/10.25585/60000745.
 	 *
-	 * @return the award ID, if present.
+	 * @return the grant ID, if present.
 	 */
-	public Optional<String> getAwardID() {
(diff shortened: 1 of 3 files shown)


src/us/kbase/workspace/database/provenance/FundingReference.java
```diff
@@ -5,83 +5,69 @@
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Optional;
-import us.kbase.workspace.database.Util;
 
 /**
- * A funding reference for a resource, including the funding body and the grant
- * awarded.
+ * A funding reference for a resource, including the funding body and the grant.
```
MrCreosote, line 10: Just so I understand this change - it seems like it's tightening the requirements for who can be a funder. With the old funderID and funderName, an individual could be a funder, but now only orgs are allowed. Similarly, `award` seems more general than `grant`, and so it seems some funding sources can no longer be represented

src/us/kbase/workspace/database/provenance/FundingReference.java
```diff
@@ -93,114 +79,94 @@ public boolean equals(Object obj) {
...
 
 	/** A builder for an {@link FundingReference}. */
 	public static class Builder {
 
-		private String funderID = null;
-		private String funderName;
-		private String awardID = null;
-		private String awardTitle = null;
-		private URL awardURL = null;
+		private Organization funder;
```
MrCreosote, line 101:
```suggestion
		private final Organization funder;
```

src/us/kbase/workspace/test/database/provenance/FundingReferenceTest.java
```diff
@@ -189,64 +158,48 @@ private void buildFundingRefFailWithError(final FundingReference.Builder builder
...
-	}
-
-	@Test
-	public void buildFailInvalidAwardURL() throws Exception {
+	public void buildFailInvalidGrantURL() throws Exception {
 		buildFundingRefFailWithError(
-				FundingReference.getBuilder(FUNDER_NAME_STRING).withAwardURL(FUNDER_ID_STRING),
-				"Illegal awardURL url '" + FUNDER_ID_STRING + "': unknown protocol: ror");
+				FundingReference.getBuilder(ORG_1).withGrantURL(GRANT_ID_STRING),
+				"Illegal grantURL url '" + GRANT_ID_STRING + "': no protocol: " + GRANT_ID_STRING);
```
MrCreosote, line 164: This test now seems to duplicate the test below, as they both test the no protocol case rather than one testing the incorrect protocol case

src/us/kbase/workspace/test/database/provenance/FundingReferenceTest.java
```diff
@@ -189,64 +158,48 @@ private void buildFundingRefFailWithError(final FundingReference.Builder builder
...
-			buildFundingRefFailWithError(
-					FundingReference.getBuilder(nullOrWs),
-					"funderName cannot be null or whitespace only");
-		}
+	public void buildFailNullFunder() throws Exception {
+		buildFundingRefFailWithError(
+				FundingReference.getBuilder(null),
+				"funder cannot be null");
+
+		buildFundingRefFailWithError(
```
MrCreosote, line 190: This test is effectively no different from the one above since there's only one `getBuilder` method

ialarmedalien: Removed redundant test.

src/us/kbase/workspace/test/database/provenance/FundingReferenceTest.java
ialarmedalien: Fixed to test a different aspect of the URL

src/us/kbase/workspace/database/provenance/FundingReference.java
ialarmedalien: Correct. This brings the CE schema terminology in closer alignment with schema.org, which will be more helpful when generating search engine indexable content for static narratives. Although sources like DataCite and Crossref use 'award' and 'grant' interchangeably, 'award' is commonly used to refer to prizes, which is not what we want to represent here.

MrCreosote: I'm a little worried that this makes things a bit too specific - the workspace has always been written with the philosophy that it's not a a KBase specific service and could work for all kinds of data. I guess it's easier to open things up later vs. tighten them down though, so for now...

src/us/kbase/workspace/database/provenance/FundingReference.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/FundingReferenceTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/FundingReferenceTest.java
MrCreosote: :+1:

src/us/kbase/workspace/test/database/provenance/FundingReferenceTest.java
```diff
@@ -189,64 +158,44 @@ private void buildFundingRefFailWithError(final FundingReference.Builder builder
...
 	@Test
-	public void buildFailNullOrWhitespaceFunderName() throws Exception {
-		for (String nullOrWs : WHITESPACE_STRINGS_WITH_NULL) {
-			buildFundingRefFailWithError(
-					FundingReference.getBuilder(nullOrWs),
-					"funderName cannot be null or whitespace only");
-		}
+	public void buildFailNullFunder() throws Exception {
+		buildFundingRefFailWithError(
+			FundingReference.getBuilder((Organization) null),
```
MrCreosote, line 187: There's no need for this cast as there isn't another `getBuilder` method with a different type, but it's only minor cruft and so NBD. Same with the one in the test below.

PR #698, opened 2023-10-18
Push metadata representation to representation layer

MrCreosote:
Regarding object information user metadata, workspace methods may

* Always return the user metadata
* Never return it
* Give the user a choice

This is because the metadata may be much bigger than the rest of the object information.

The way this is implemented is that each method has a list of MongoDB fields it asks for when querying the DB. If it doesn't want metadata, it doesn't include the field containing the metadata, and so in the returned data the metadata will be effectively null. If the method asks for metadata and there's none available, an empty map will be returned from the DB since that's what's stored in the metadata field when none is supplied when saving objects.

This is a problem for admin metadata, since older objects will have null for the admin metadata whether the field is requested or not. This means without changes, if a method requests metadata they'll get null rather than the expected emtpy map.

I first tried to fix this by passing a boolean around to the point at which the ObjectInformation class is built telling it whether to translate nulls to empty maps but that was a huge mess.

Instead, I updated ObjectInfomation to treat nulls and empty maps the same - an absence of metadata. The representation layer can then be responsible for deciding whether the absent metadata representation should be null (for no metadata requested) or an empty map.

Note that none of the API tests that test the representation layer needed changes.


src/us/kbase/workspace/database/ObjectInformation.java
```diff
@@ -149,12 +149,22 @@ public Optional<UncheckedUserMetadata> getUserMetaData() {
 		return Optional.ofNullable(usermeta);
 	}
 	
-	/** Returns the user supplied and automatically generated metadata for the object as a map
-	 * or null if no user metadata was provided.
+	/** Returns the user supplied and automatically generated metadata for the object as a map.
+	 * 
+	 * @param nullIfEmpty return null rather than an empty map if there is no metadata available.
```
ialarmedalien, line 154: does the `@return` description below need to be updated?

MrCreosote: Fixed

ialarmedalien (approved):

PR #707, opened 2023-10-25
Add API translation method for setting admin object metadata

MrCreosote:


src/us/kbase/workspace/AlterAdminObjectMetadataParams.java
MrCreosote, line 1: This and the next class are compiled from an updated spec which isn't included in order to split the changes up into smaller chunks.

MrCreosote:
*&*)&_(*)^ codecov

ialarmedalien (approved):

PR #722, opened 2024-03-26
Gradle pt 1: tests + coverage

MrCreosote:


src/us/kbase/typedobj/db/KidlUtil.java
```diff
@@ -1,8 +1,5 @@
 package us.kbase.typedobj.db;
 
-import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
```
MrCreosote, line 4: Adding test dependencies in implementation code is rather naughty

MrCreosote:
Sigh, codecov is saying the uploaded reports are bad

.classpath
```diff
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="src" path="server_scripts"/>
 	<classpathentry excluding=".gitignore" kind="src" path="src"/>
-	<classpathentry kind="src" path="performanceanddevelopment"/>
```
Xiangs18, line 5: 1. do we need to keep these two folders as cp are removed?
2. do we need to keep .classpath file as we switch build tool from Ant to Gradle?

MrCreosote: 1. Yes, the code in there is still useful as a historical resource. I just removed it from the classpath to stop eclipse yelling at me about it
2. That'll happen in one of the last commits

.codecov.yml
```diff
@@ -26,5 +26,4 @@ comment:
   require_changes: no
 
 ignore:
-  - "build"
```
Xiangs18, line 29: ignore build folder?

MrCreosote: Stop ignoring the build folder, which is where jacoco puts the code coverage data

Xiangs18: Why it was ignored before then? Now is removed to boost test coverage?

MrCreosote: I'm not sure, it might be just a codecov default. There was no build directory before so it didn't matter, but Gradle puts everything there so codecov needs to look there to find the coverage data

Xiangs18: 👍

.classpath
Xiangs18: 👍

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+	File file = new File("$buildDir/download/${name}.jar")
+	file.parentFile.mkdirs()
+	if (!file.exists()) {
+		new URL(url).withInputStream { downloadStream ->
+			file.withOutputStream { fileOut ->
+				fileOut << downloadStream
+			}
+		}
+	}
+	files(file.absolutePath)
```
Xiangs18, line 144: What is this step for?

MrCreosote: https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#files-java.lang.Object...-

I believe it's that. I'm still a bit of a Gradle beginner so my understanding here is that it's converting a java `File` into a type that the Gradle `implementation` understands

MrCreosote: This whole function is something I just found and copied, to be perfectly honest. It works though

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+			include '**/*.spec'
+			include '**/*instance.*'
+			include '**/*.instance*'
+			include '**/*.html'
+			include '**/*.css'
+			include '**/*.gif'
+			include '**/*.js'
+			include '**/*.png'
+			include '**/*.txt'
+			include '**/*.weirdsuffix'
```
Xiangs18, line 112: Are these for future commits? There are not many files in the `test` folder

MrCreosote: These are all current files currently used in tests, but all the tests are under the src folder

Xiangs18: I assume they this part will be deleted once we move test folder out?

MrCreosote: yeah, I'll try to set things up like auth so we don't need the `sourceSets` section at all, but that comes last

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+}
+
+test {
+	finalizedBy jacocoTestReport
+}
+
+task testQuick(type: Test) {
+	// for Gradle 9.0 compatibility
+	testClassesDirs = testing.suites.test.sources.output.classesDirs
+	classpath = testing.suites.test.sources.runtimeClasspath
```
Xiangs18, line 74: Are those default env vars? They are not used elsewhere in this script.

MrCreosote: They're going to be required in Gradle 9 for non-default test tasks to work

Xiangs18: 👍

build.gradle
Xiangs18: 👍

Xiangs18 (approved):
LGTM! @ialarmedalien should take a look as well :)

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+// TODO TEST avoid early configuration, see
+//           https://docs.gradle.org/current/userguide/task_configuration_avoidance.html
+//           need to avoid withType as well apparently?
+
+plugins {
+	id 'java'
+	id 'war'
+	id 'jacoco'
+	id 'org.ajoberstar.grgit' version '4.1.1'
+	id 'com.github.johnrengelman.shadow' version '8.1.1'
```
ialarmedalien, line 15: what are these two for?

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+	main {
+		java {
+			srcDirs = ["src"]
+			exclude '**/test/**'
+		}
+	}
+	test {
+		java {
+			srcDirs = ["src"]
+			include '**/test/**'
```
ialarmedalien, line 98: won't this automatically be included?

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+	implementation 'info.picocli:picocli:4.6.1'
+	implementation 'org.mongodb:mongodb-driver-core:4.11.1'
+	implementation 'org.mongodb:mongodb-driver-sync:4.11.1'
+	implementation 'org.mongodb:bson-record-codec:4.11.1'
+	implementation 'org.mongodb:bson:4.11.1'
+	implementation 'org.slf4j:slf4j-api:1.7.30'
+	implementation 'ch.qos.logback:logback-classic:1.1.2'
+	implementation 'com.google.guava:guava:14.0.1'
+	implementation 'com.github.ben-manes.caffeine:caffeine:2.9.3'
+	implementation 'org.apache.kafka:kafka-clients:2.1.0'
```
ialarmedalien, line 175: can you alphabetise this list?

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+}
+
+configurations {
+	// can't directly access testImplementation, so extend and access
+	testimpl.extendsFrom testImplementation
+}
+
+def fromURL = { url, name ->
+	File file = new File("$buildDir/download/${name}.jar")
+	file.parentFile.mkdirs()
```
ialarmedalien, line 136: does this check for directory existence first?

build.gradle
MrCreosote: First one will be for pulling the git commit for building into the server, so we don't have to rely on kb-sdk compile. Second one will be for building a shadow jar, which Gradle can't do out of the box

build.gradle
MrCreosote: It works regardless of whether the build directory exists or not, so presumably it's a `mkdir -p` situation.

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+	// this is OOOOOOLD. But that probably means updating java_common
+	implementation 'org.eclipse.jetty.aggregate:jetty-all:7.0.0.v20091005'
+	implementation 'javax.servlet:servlet-api:2.5'
+	implementation 'javax.annotation:javax.annotation-api:1.3.2'
+	// Syslog4j 0.9.46 doesn't appear to be available on Maven. It apparently lives in 
+	// a JetBrains artifact server, but that's too much trouble and there's only one version there
+	// anyway.
+	// https://mvnrepository.com/artifact/org.jetbrains/syslog4j/0.9.46
+	// Need to rework the java common logger to not use syslog4j at all since it's abandonware
+	// and has a ton of CVEs, even in the newer versions.
```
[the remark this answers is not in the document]
MrCreosote: In the next commit I get rid of the JNA dependency since it prevents the server from working in a shadowjar, so I don't actually want to pull it from a repo since that'll force the JNA dependency. Plus it's no harder to pull it from jars. Plus plus we need to get rid of sylog4j anyway so a janky pulldown makes that more obvious. I'll update the comment to be more accurate along those lines

build.gradle
MrCreosote: done. I put all the jars repo deps first and then alphabetized each subsection. Re the subsections see comment below

build.gradle
```diff
@@ -0,0 +1,248 @@
...
+	implementation fromURL(
+		'https://github.com/kbase/jars/raw/master/lib/jars/syslog4j/syslog4j-0.9.46.jar',
+		'syslog4j-0.9.46'
+	)
+	// needed for syslog4j
+	implementation 'net.java.dev.jna:jna:3.4.0'
```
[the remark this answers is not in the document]
MrCreosote: The subsections and comments like this are for cases where Gradle can't figure out the dependency tree, e.g.

* the java common section contains the dependencies for the kbase-common jar, which has no manifest
* Same for the shock client section
* amazon SDK doesn't automatically pull in the url client

As such I want to group these sections together to make it clear where the dependencies are coming from and why they're in the list.

build.gradle
MrCreosote: They're both used in auth2 if you want to see examples

ialarmedalien (approved):

PR #723, opened 2024-03-28
Remove JNA

MrCreosote:
JNA doesn't work in a shadow jar, which we'll need for workspace test rigs in other repos


ialarmedalien (approved):

Xiangs18:
I am not that familiar with syslog4j. Is it similar to Log4j?

MrCreosote:
No, it's used for sending log information to the linux syslog daemon as opposed to being a logging framework like log4j or slf4j

Xiangs18:
Could you pleas explain more? Still try to understand why a simple logging framework like log4j is not in favor here.

MrCreosote:
The system has a logging framework - it uses slf4j. slf4j does not provide code for logging to the syslog daemon. Syslog4j does, and so it was used as that was the requirement at the time

Xiangs18 (approved):
LGTM!

PR #724, opened 2024-03-28
Build docs

MrCreosote:


workspace.spec
```diff
@@ -980,7 +980,7 @@ module Workspace {
 		UnspecifiedObject data - The object's data.
 		object_metadata metadata - Metadata for object retrieved/
 
-		@deprecated Workspaces.ObjectData
+		@deprecated Workspace.ObjectData
```
MrCreosote, line 983: oops

build.gradle
```diff
@@ -32,6 +34,54 @@ compileJava {
...
+	 */
+	failOnError = false
+	options {
+		links "https://docs.oracle.com/en/java/javase/11/docs/api/"
+		links "https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-core/2.9.9/"
+		links "https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/2.9.9/"
+	}
+	include "**/workspace/*.java"
+	exclude "**/workspace/WorkspaceServer.java"
+	include "**/common/service/Tuple*.java"
```
MrCreosote, line 60: @ialarmedalien I know you laboriously copied over all the options from ant in your PR, but frankly I have no idea why they were there in the first place or if they're needed. The docs without them look fine to me so in the interests of simplicity I left them out. If we find we need them in the future it's easy enough to put them back in

build.gradle
```diff
@@ -32,6 +34,54 @@ compileJava {
 	options.release = 11
 }
 
+javadoc {
```
Xiangs18, line 37: What is the purpose of creating javadoc here?
I have no problem understanding `task buildDocs`, but not sure why it depends on javadoc.

MrCreosote: Because the purpose of the task is to build all the docs, which means the javadoc task need to run as well. The dependency forces that to happen

Xiangs18: 👍

build.gradle
```diff
@@ -32,6 +34,54 @@ compileJava {
...
+	 */
+	/* TODO DOCS the current sdk documenation looks like invalid html to javadoc
+	 *           need to go through and remove
+	 *           also some spots with < / > that need to be wrapped with {@code } in internal code
+	 */
+	failOnError = false
+	options {
+		links "https://docs.oracle.com/en/java/javase/11/docs/api/"
+		links "https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-core/2.9.9/"
+		links "https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/2.9.9/"
```
Xiangs18, line 56: May I know why we need to include them here?

MrCreosote: you know what, we don't actually need the jackson links. When the java common docs were built into the workspace docs they were needed since they had references to jackson, but if we do it the right way and host the java docs then these links will be needed in java common, not here

Xiangs18 (approved):
LGTM! 
@ialarmedalien should take a look as well!

build.gradle
ialarmedalien: Not sure why you didn't just reuse the code that I'd already written, esp since I laboriously tested it to ensure that it generated the same stuff. Also IMO it would have made more sense to start by building the jars, then the wars, then the docs, rather than going at things the other way around.

workspace.spec
ialarmedalien: ah, it's only the spec... not a proper code file!

ialarmedalien (approved):

build.gradle
MrCreosote: I need the docs for the war and the shadow jar, which is why I'm going them first

PR #726, opened 2024-04-04
Add shadow jar and workspace controller that starts said jar

MrCreosote:
Tested this setup in the Groups service, replacing the workspace controller there

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,7 @@ var BUILD_DOC_ROOT = "$buildDir/docs/"
 var BUILD_JAVA_DOC_DIR = "$BUILD_DOC_ROOT/javadoc"
 var BUILD_OTHER_DOC_DIR = "$BUILD_DOC_ROOT/otherdoc/"
 // This is where the DocServer will look for docs, if it changes it needs to change
-// in deloy.cfg as well
+// in deploy.cfg as well
 // TODO DOCS just get rid of configuring this in the DocServer, make things hardcoded
 //           or pass in constructor when we subume into a Jersey style service
 var IN_JAR_DOCS_DIR = "/server_docs"
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,14 @@ plugins {
 	id 'com.github.johnrengelman.shadow' version '8.1.1'
 }
 
-var buildDocsDir = "$buildDir/docs/otherdoc/"
+var BUILD_DOC_ROOT = "$buildDir/docs/"
+var BUILD_JAVA_DOC_DIR = "$BUILD_DOC_ROOT/javadoc"
+var BUILD_OTHER_DOC_DIR = "$BUILD_DOC_ROOT/otherdoc/"
+// This is where the DocServer will look for docs, if it changes it needs to change
+// in deloy.cfg as well
+// TODO DOCS just get rid of configuring this in the DocServer, make things hardcoded
+//           or pass in constructor when we subume into a Jersey style service
+var IN_JAR_DOCS_DIR = "/server_docs"
 
 // TODO NOW test shadow jar works in groups - need test controller - commit 2
 // TODO NOW get docserver working in WAR and test in docker-compose <- edit to start test server - commit 3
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -267,10 +294,9 @@ dependencies {
 	)
 	testImplementation 'com.arangodb:arangodb-java-driver:6.7.2'
 	testImplementation 'com.github.zafarkhaja:java-semver:0.9.0'
-	testImplementation 'org.hamcrest:hamcrest-core:1.3'
-	testImplementation 'commons-lang:commons-lang:2.4'
```
MrCreosote, line 271: Turns out this isn't needed

Xiangs18: commons-lang is not needed?

build.gradle
```diff
@@ -132,6 +139,26 @@ task jacocoTestQuickReport(type: JacocoReport, dependsOn: testQuick) {
...
+	// we don't want to package the auth shadow jar, so no test runtime classpath
+	configurations = [project.configurations.runtimeClasspath]
+
+	enableRelocation true
+	relocationPrefix 'us.kbase.workspace.shadow'
+	
+	mergeServiceFiles()
+	
+	from(BUILD_JAVA_DOC_DIR) { into "$IN_JAR_DOCS_DIR/javadoc" }
+	from(BUILD_OTHER_DOC_DIR) { into IN_JAR_DOCS_DIR }
```
Xiangs18, line 159: Are these two lines just copy files into /server_docs?

MrCreosote: yep

Xiangs18: so they are in shadow jar. Why we need to do this? Or is just for documentation.

MrCreosote: So that if someone runs a test against the shadowjar that touches the /docs endpoint for some reason things work as expected

Xiangs18: how are /docs and /server_docs connected? It somehow knows to look into /server_docs when a test touches /docs endpoint?

Xiangs18: 👍

Xiangs18 (approved):
LGTM!
@ialarmedalien should also take a look!

build.gradle
```diff
@@ -15,7 +15,14 @@ plugins {
 	id 'com.github.johnrengelman.shadow' version '8.1.1'
 }
 
-var buildDocsDir = "$buildDir/docs/otherdoc/"
+var BUILD_DOC_ROOT = "$buildDir/docs/"
+var BUILD_JAVA_DOC_DIR = "$BUILD_DOC_ROOT/javadoc"
+var BUILD_OTHER_DOC_DIR = "$BUILD_DOC_ROOT/otherdoc/"
+// This is where the DocServer will look for docs, if it changes it needs to change
+// in deloy.cfg as well
```
briehl, line 22: typo: `deploy.cfg`

src/us/kbase/workspace/test/controllers/workspace/WorkspaceController.java
```diff
@@ -0,0 +1,115 @@
...
+		final ProcessBuilder servpb = new ProcessBuilder(command)
+				.redirectErrorStream(true)
+				.redirectOutput(tempDir.resolve("workspace.log").toFile());
+
+		final Map<String, String> env = servpb.environment();
+		env.put("KB_DEPLOYMENT_CONFIG", deployCfg.toString());
+
+		workspace = servpb.start();
+		// TODO TEST add periodic check w/ exponential backoff
+		Thread.sleep(5000);
```
briehl, line 71: Is this just to make sure that connections get made properly? Are there asynchronous things (like  `servpb.start()`) that go off in this time that are more of a pain to monitor?

briehl (approved):
Gotta be honest, I'm not entirely sure how to review here. I looked at the shadow jar docs (https://imperceptiblethoughts.com/shadow/) and the gradle stuff looks fine to me.

So here's the summary if I'm reading it right.

`build.gradle`'s updated to make a shadow jar - a single jar containing the whole workspace service + docs + the new `WorkspaceController`.

`WorkspaceController` can then be invoked to stand up a temporary Workspace service to... do stuff? Does it run its own tests? Does that shadow jar get included with other Java-based services to run a temporary workspace a little more seamlessly than running, say, a Docker image?

Anyway, that looks like (from some other test module) you can just go:
```
final WorkspaceController ws = new WorkspaceController(params);
... do test stuff ...
ws.destroy();
```
which is kinda awesome.

Anyway, if that's the intent, this all seems to hit all the buttons and link together. Hitting approve, but there's a typo and a question there, at your leisure.

MrCreosote:
WorkspaceController can then be invoked to stand up a temporary Workspace service to... do stuff? Does it run its own tests? Does that shadow jar get included with other Java-based services to run a temporary workspace a little more seamlessly than running, say, a Docker image?

It's for cases like this: https://github.com/kbase/groups/blob/master/src/us/kbase/test/groups/integration/ServiceIntegrationTest.java#L124-L132

With the shadowjar, you no longer need to list all the dependent jars: https://github.com/kbase/groups/blob/master/src/us/kbase/test/groups/controllers/workspace/wsjars

So you just add that one jar to your classpath and can easily start up a workspace to test against. 

Eventually I want the workspace repo to publish the shadow jar via jitpack but that's a bit farther down the road

src/us/kbase/workspace/test/controllers/workspace/WorkspaceController.java
MrCreosote: The workspace takes a second or two to start up after `ProcessBuilder.start()` is called to check all the connections to other services. This just waits until it should be ready to accept connections.

MrCreosote:
> which is kinda awesome.

and it's **totally** awesome, thank you very much

src/us/kbase/workspace/test/controllers/workspace/WorkspaceController.java
MrCreosote: I could do a microoptimization by calling workspace.ver() in a loop waiting for it to come up as per the TODO but it doesn't really seem worth if for now

build.gradle
MrCreosote: Fixed

PR #728, opened 2024-04-10
Build docker image with gradle

MrCreosote:


Dockerfile
```diff
@@ -2,15 +2,27 @@ FROM eclipse-temurin:11-jdk as build
 
 WORKDIR /tmp
```
Xiangs18, line 3: this line could be removed as well?

MrCreosote: No, the ./gradlew war command depends on it

Dockerfile
```diff
@@ -2,15 +2,27 @@ FROM eclipse-temurin:11-jdk as build
...
+RUN ./gradlew dependencies
+
+# Now build the code
+COPY workspace.spec /tmp/workspace/workspace.spec
+COPY deployment/ /tmp/workspace/deployment/
+COPY docshtml /tmp/workspace/docshtml/
+COPY docsource /tmp/workspace/docsource/
+COPY lib /tmp/workspace/lib/
+COPY src /tmp/workspace/src/
+COPY war /tmp/workspace/war/
```
Xiangs18, line 23: Already set `WORKDIR /tmp/workspace`, maybe we can simply `/tmp/workspace` as `.` when doing copy

Dockerfile
MrCreosote: Fixed

Dockerfile
MrCreosote: I would rather keep paths absolute and explicit when possible. `gradlew` needs to run in the project directory AIUI

MrCreosote:
Codecov report is here: https://github.com/kbase/workspace_deluxe/pull/729#issuecomment-2048234433

MrCreosote:
It seems like it's no longer possible to run codecov from a fork

Xiangs18 (approved):
LGTM!

briehl (approved):

PR #734, opened 2024-04-23
Move source files to Gradle default locations

MrCreosote:


src/test/java/us/kbase/test/common/MapBuilder.java
```diff
@@ -1,4 +1,4 @@
-package us.kbase.common.test;
+package us.kbase.test.common;
```
ialarmedalien, line 1: OK, so the `test` directory has been moved up a level instead of having a test dir under each of the three subdirectories. 👍

ialarmedalien (approved):

Xiangs18 (approved):
LGTM!

PR #738, opened 2024-05-01
Add a worksapce client module

MrCreosote:
The purpose of this module is to allow for serving the workspace client jar out of jitpack.io, which can only serve 1 jar per module.

I figured this approach was better than making a separate repo and having to sync over the spec, make sure the github releases were equivalent & up to date, etc.

The drawback is the duplicate generated code, but a 2nd repo would have that drawback as well.


client/src/main/java/us/kbase/common/service/Tuple11.java
MrCreosote, line 1: Everything in client/src is SDK generated

Xiangs18:
how to make sure generated code is up to date?

MrCreosote:
./gradlew sdkCompile compiles the generated code for both the client and server subprojects

Xiangs18 (approved):

MrCreosote:
It's 100% clunky. It's just the least clunky option I'm aware of

PR #750, opened 2025-06-16
SECURITY-4: Update Dependabot.yml

Xiangs18:


MrCreosote:
Looks like the tests are failing

MrCreosote:
Changes look ok other than the same note as prior reviews re the time not really being necessary, but fine

Xiangs18:
The test failures are not related to this update.
https://github.com/kbase/workspace_deluxe/pull/749#issuecomment-2683530539

MrCreosote:
Are the handle service and sample service still on 3.7? If they've been updated, try (in a different PR) to update the ubuntu version and the python version to see if the test pass. If they do, we can merge that and then rerun the tests here.

If not or the python version is still too low, then we should convert the workspace repo to `uv` and use it to install the necessary python version. There's an example here https://github.com/kbase/kb_sdk_plus/pull/81

MrCreosote:
Also if the services' dependencies have been updated they should be synced here come to think of it

Xiangs18:
> Looks like the tests are failing

ws test fail fixed by this pr: https://github.com/kbase/workspace_deluxe/pull/751

MrCreosote (approved):

PR #751, opened 2025-06-16
Update test.yml and fix failed tests

Xiangs18:


Xiangs18:
@MrCreosote Any idea what might be causing the Sample Service connection refused?

https://github.com/kbase/workspace_deluxe/actions/runs/15863570581/job/44725978140?pr=751#step:6:832

MrCreosote:
Not off the top of my head, you'll need to check the sample service logs

.github/workflows/test.yml
```diff
@@ -19,12 +19,12 @@ on:
 jobs:
 
   workspace_container_tests:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
MrCreosote: Please pin the ubuntu version

.github/workflows/test.yml
```diff
@@ -57,18 +57,18 @@ jobs:
...
           distribution: 'temurin'
           java-version: ${{matrix.java}}
 
-      - name: Set up Python 3.7  # sample service is on 3.7, handle service is on sdkbase2.latest
-        uses: actions/setup-python@v3
+      - name: Set up Python
+        uses: actions/setup-python@v5
         with:
-          python-version: "3.7"
+          python-version: "3.9.19"
```
MrCreosote, line 71: Both sample service and handle service are on 3.9.19?

.github/workflows/test.yml
```diff
@@ -108,7 +108,7 @@ jobs:
           cd ../..
 
           # set up sample service
-          export SAMPLE_COMMIT=6813fb148e95db2b11db6eea04f4d1d45cbb7119
+          export SAMPLE_COMMIT=cb8129dded67ceca57cc175a1c7c357c56e466b3
```
MrCreosote: What's the difference between these commits?

python_dependencies/Pipfile
```diff
@@ -18,20 +18,20 @@ uwsgi = "==2.0.18"
...
 pygtrie = "==2.3.2"
 jinja2 = "==2.11.2"
-gevent = "==20.5.0"
-gunicorn = "==20.0.4"
+gevent = "==24.11.1"
+gunicorn = "==20.1.0"
 kafka-python = "==2.0.1"
-greenlet = "==0.4.16"
+greenlet = "==3.1.1"
 typing-extensions = "*"
```
MrCreosote: These are all the same as what's current in sample service?

.github/workflows/test.yml
MrCreosote: Should be pinned below as well, not sure why it isn't

Xiangs18: All CI tests across core and supporting services use ubuntu-latest, as they should - code should work with the latest environment rather than being tied to a specific OS version.

.github/workflows/test.yml
Xiangs18: yes

python_dependencies/Pipfile
Xiangs18: yes

.github/workflows/test.yml
Xiangs18: cb8 is the latest commit on the develop branch of the Sample Service, authored by me as part of the GitHub Actions workflow update. 6813 is an older commit made by Dakota back in 2022.

.github/workflows/test.yml
MrCreosote: No, tests should be pinned to a specific version so they don't start failing unexpectedly when versions silently change. This is the same as pinning any other dependency. 

There's no guarantee code will work with newer os or dependency versions, since there's no way for the code to anticipate what changes might be made

.github/workflows/test.yml
MrCreosote: Ok, we should be testing based on what's in production vs. what's on develop

python_dependencies/Pipfile
MrCreosote: Looking at https://github.com/kbase/sample_service/blob/develop/Pipfile:

apscheduler = "==3.9.1"
cacheout = "==0.13.1"
pyyaml = "==6.0"
python-ranges = "==0.2.0"
pygtrie = "==2.4.2"
jinja2 = "==3.1.1"
kafka-python = "==2.0.2"
types-pyyaml = "==6.0.7"
types-requests = "==2.27.20"
shell = "==1.0.1"

.. are different. Am I looking in the wrong place?

.github/workflows/test.yml
MrCreosote: :+1:

.github/workflows/test.yml
MrCreosote: Wait - is the issue that GHA no longer supports the version of python prod sample service uses?

.github/workflows/test.yml
Xiangs18: No, the code should be compatible with modern, up-to-date environments. If it breaks with a newer version of Ubuntu, we should fix it promptly - this prevents technical debt from accumulating and being offloaded onto a single person later. That’s the whole point of initiating the Security White Paper project in the first place.

MrCreosote: It can't be. Environments make breaking changes, and we don't have the personnel to immediately be able to stop what they're doing and fix broken tests and dependencies. That's the point of the dashboard we're going to make - to triage which repos need work the most.

The approach you're suggesting would mean we should unpin all our dependencies, not just ubuntu, which we're clearly not going to do.

python_dependencies/Pipfile
Xiangs18: The dependency updates I made align with those currently used in the Sample Service, except for apscheduler. I can downgrade it to version 3.9.1 if preferred, but the original version is clearly causing issues.

MrCreosote: So why is the pipfile in develop different?

.github/workflows/test.yml
Xiangs18: I'm only suggesting the use of ubuntu-latest in test CI - not unpinning all our dependencies. If we're enabling Dependabot to keep our dependencies up to date, why shouldn't we do the same for the ubuntu version? These two actions should be consistent, not contradictory. Besides, Ubuntu updates far more slowly than most Python packages.

python_dependencies/Pipfile
Xiangs18: Are they the same to begin with?

MrCreosote: No, that's why I'm asking. The Pipfile here isn't synced with the Pipfile in the sample service develop branch, so we're not testing against the same thing. 

...and actually, we really should be testing against what's in production, not what's on develop.

.github/workflows/test.yml
MrCreosote: >  If we're enabling Dependabot to keep our dependencies up to date, why shouldn't we do the same for the CI environment? 

The difference for our dependencies is that Dependabot just creates an update PR; it doesn't perform the update.

If we can make Dependabot make PRs to update Ubuntu in GHA yml files I'm all for that. I'm against allowing GHA to update Ubuntu without us explicitly telling it to do so

> These two actions should be consistent, not contradictory. 

If we allow Ubuntu to update whenever that would be contradictory with how we manage dependencies

python_dependencies/Pipfile
Xiangs18: The Sample Service is still at version 0.2.5 in ci, while the GitHub Actions workflows and Python updates are in version 0.2.7. on develop. Shouldn’t we deploy 0.2.7 to ci?

Xiangs18: If we're testing against what's currently in production, this PR might be overkill for now - but it will definitely be needed once version 0.2.7 is deployed to ci.

MrCreosote: From what I recall, there's some problem we don't understand with the current version of the sample service that caused sample lookups (or something, I don't quite remember) to become unusably slow, so we rolled back. We'd need to figure that out before we deployed to CI or anywhere else.

What are the minimum changes required in this PR to get tests to pass?

Xiangs18: > What are the minimum changes required in this PR to get tests to pass?

May I know which version of the Sample Service you'd like to test against?

MrCreosote: According to this, even 0.2.5 is broken: https://github.com/kbase/sample_service/issues/481

MrCreosote: Whatever is in production, which I'm guessing based on the above ticket is 0.2.4. 

If we can't test that because of GHA constraints, what are the minimal set of deviations required to get tests to pass in this PR?

Xiangs18: currently is 0.2.5 in ci.

If we're testing against version 0.2.4/0.2.5, my understanding is that we just need to manually install Python 3.7 in GHA

MrCreosote: I'm fine with a deviation to 3.9 in GHA, that seems low risk, assuming it works

Xiangs18: So, just to confirm, you're aiming for the following:

1. Use Python 3.9 in GitHub Actions
2. Continue testing the original Sample Service commit (6813fb1)
3. Make only the minimal changes necessary to get all tests passing

MrCreosote: > Use Python 3.9 in GitHub Actions

If that's the closest we can get to what the sample service is running in prod, yeah. What is the sample service running in prod? What about the handle service?

> Continue testing the original Sample Service commit (6813fb1)

What commit is running in prod?

> Make only the minimal changes necessary to get all tests passing

I would say make the changes necessary to have the tests mirroring prod configurations of the sample and handle services. If those configurations are impossible to test in a reasonably standard way in GHA (e.g. no manual python installs), let's discuss the deviations necessary. Using python 3.9 is fine and doesn't need further discussion

Xiangs18: BTW, the current test setup in the workspace GHA is suboptimal. Each service should run in its own container rather than being pulled from github and included as a submodule.

Xiangs18: The Sample Service is running version 0.2.5 (commit b7e68), authored by Dakoda in 2022, with Python 3.7 in ci.
The Handle Service is running version 1.0.7 (commit 2213d), authored by me in 2024, with Python 3.9.19 in ci.

MrCreosote: What's in prod?

Xiangs18:
```
curl -d '{"id":42,"method":"SampleService.status","params":[]}' https://kbase.us/services/sampleservice
```
{"result": [{"state": "OK", "message": "", "version": "0.2.1", "git_url": "git@github.com:charleshtrenholm/sample_service.git", "git_commit_hash": "aa0303c4d5d4aa1e3cf0c9ecb36366fb1127f93e", "servertime": 1751497609859}], "id": 42}


The Handle service's prod is same as ci

MrCreosote: Wow, old.

Ok, so what's the difference between the commit we're testing and that commit?

Xiangs18: Too many commits are being referenced - can you specify exactly which two you mean?

MrCreosote: The commit in the comment 3 comments up and the commit that's currently checked into the test.yml file in develop

Xiangs18: Neither of the commits was authored by me, but I believe the difference is between versions 0.2.1 and 0.1.1. You can find more details in the Sample Service [release notes](https://github.com/kbase/sample_service/blob/develop/RELEASE_NOTES.md#021).

Screenshot 2025-07-02 at 4 31 30 PM

MrCreosote: Did you confirm the commits line up with those tags?

Xiangs18: https://github.com/kbase/sample_service/releases/tag/0.1.1

MrCreosote: What about 0.2.1?

Xiangs18: https://github.com/kbase/sample_service/releases/tag/0.2.1

MrCreosote: Ok, so looks like the thing to do is sync the sample service configuration here to as close as possible to 0.2.1

Xiangs18: What about the handle service? Its current checkout commit is also out of sync with production.

MrCreosote: what we're testing is older than what's in production, right? If so yeah, sync it up

Xiangs18: old, 1.0.4 vs 1.0.7 in prod

Xiangs18:
Self-note: The file /lib/biokbase/log.py already exists in Handle Service v1.0.7, so there's no need to retrieve it separately from https://raw.githubusercontent.com/kbase/sdkbase2 - it's the same file.

.github/workflows/test.yml
Xiangs18: updated SAMPLE_COMMIT. Now test based on what's in prod

.github/workflows/test.yml
Xiangs18: Done

python_dependencies/Pipfile
Xiangs18: Now align as closely as possible with the Sample Service and Handle Service in production. For any deviations in dependencies within the Sample Service, refer to the commit messages below for justification.

.github/workflows/test.yml
MrCreosote: :+1:

python_dependencies/Pipfile
MrCreosote: Can you add those notes as comments into the pipfile please?

python_dependencies/Pipfile
```diff
@@ -11,27 +11,31 @@ jsonrpcbase = "==0.2.0"
...
+gevent = "==24.11.1"
 gunicorn = "==20.0.4"
 kafka-python = "==2.0.1"
-greenlet = "==0.4.16"
+greenlet = "==3.1.1"
 typing-extensions = "*"
+types-pyyaml = "*"
+types-requests = "*"
+pipenv = "*"
+shell = "*"
```
MrCreosote, line 42: For the unpinned deps in the pipfile, are they unpinned in their respective repos as well?

Xiangs18: you mean in the sample service v0.2.1?

MrCreosote: There's one in handle service as well

Xiangs18: These four are from Sample Service v0.2.1. I just removed `configparser = "*"`, which was not present in Handle Service v1.0.7 pipfile anymore.

MrCreosote: Are they unpinned there?

Xiangs18: yes, they are unpinned in Sample Service v0.2.1

python_dependencies/Pipfile
MrCreosote: Looks like HS is on requests = "==2.31.0"

Xiangs18: By `notes`, do you mean my comment above?

Xiangs18: > Looks like HS is on requests = "==2.31.0"

Is a joint dependency. SS is on requests = "==2.22.0", so i did not modify.

python_dependencies/Pipfile
MrCreosote: Looks like they're pinned now and pipenv is removed, so this is the best we can do I guess. If the tests start failing we can look into the sample service container and figure out what it's actually running.

MrCreosote: :+1:

python_dependencies/Pipfile
MrCreosote: > By notes, do you mean my comment above? 

right, it'll be a lot easier to understand why the versions are the way they are for future folks if the justifications are right by the versions

.github/workflows/test.yml
MrCreosote: :+1:

python_dependencies/Pipfile
MrCreosote: Looks like deviations are

apscheduler = "==3.6.3"
pyyaml = "==5.4"
pint = "==0.10.1"
gevent = "==20.5.0"
greenlet = "==0.4.16"

other than the unpinned stuff

Xiangs18: jsonschema should not be included in the list of deviations.

MrCreosote: Fixed. Now I'm not sure if I left something out...

Xiangs18: I've double-checked - it should be correct now.

Xiangs18: > Can you add those notes as comments into the pipfile please?

Done

MrCreosote: Please just put the justifications for the deviations directly in the pipfile

Xiangs18: added

MrCreosote (approved):

